PR #12, opened 2014-11-28
fix panic from stringToBytes on addr with missing argument

ebuchman:
re https://github.com/jbenet/go-multiaddr/pull/11 (used the wrong length check there, believe this one is correct, with tests to prove it, sorry it took so long)

The test commit panics. This should be caught, especially since I can trigger it by editing `~/.go-ipfs/config`, for example by changing one of my swarm listening addresses there to `/ip4` or "/ip4/0.0.0.0/tcp/" and running `ipfs daemon`.  

Also, wondering, why are the http/s commented out?

### DIFF
--- a/codec.go
+++ b/codec.go
@@ -32,6 +32,9 @@ func stringToBytes(s string) ([]byte, error) {
 		sp = sp[1:]
 
 		if p.Size > 0 {
+			if len(sp) == 0 {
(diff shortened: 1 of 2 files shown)


Stebalien:
This has now been fixed and sufficient tests have been added. (sorry for leaving this hanging).

PR #13, opened 2015-08-29
Add Tor .onion address support and unit test cases

david415:
This should be reviewed... Let's add Tor onion addressing to go-multiaddr.

Cheers,


leif:
It's worth considering that in the future there will be next-gen onion services, which will definitely have longer names (256-bit keys instead of 80-bit hashes of keys) and will likely also use .onion and which might even use a different base32 encoding... but I think this strict validation is good for now and can easily be extended to support next-gen addresses later.

protocols.go
```diff
@@ -46,6 +47,7 @@ var Protocols = []Protocol{
 	Protocol{P_IP6, 128, "ip6", CodeToVarint(P_IP6)},
 	// these require varint:
 	Protocol{P_SCTP, 16, "sctp", CodeToVarint(P_SCTP)},
+	Protocol{P_ONION, 10, "onion", CodeToVarint(P_ONION)},
```
jbenet: `10` seems off? shouldn't this be `80` ? (sorry for being inconsistent, the multihash standard is bytes for compactness, this is in bits. arguably should be bytes).

codec.go
```diff
@@ -165,6 +166,36 @@ func addressStringToBytes(p Protocol, s string) ([]byte, error) {
 		binary.BigEndian.PutUint16(b, uint16(i))
 		return b, nil
 
+	case P_ONION:
+		addr := strings.Split(s, ":")
+		if len(addr) != 2 {
+			return nil, fmt.Errorf("failed to parse %s addr: %s does not contain a port number.", p.Name, s)
```
jbenet: Should this default to `:80` ?

codec.go
```diff
@@ -165,6 +166,36 @@ func addressStringToBytes(p Protocol, s string) ([]byte, error) {
...
+		if err != nil {
+			return nil, fmt.Errorf("failed to decode base32 %s addr: %s %s", p.Name, s, err)
+		}
+
+		// onion port number
+		i, err := strconv.Atoi(addr[1])
+		if err != nil {
+			return nil, fmt.Errorf("failed to parse %s addr: %s", p.Name, err)
+		}
+		if i >= 65536 {
```
jbenet: also `i < 0`? (or whatever the minimum port is)

multiaddr_test.go
```diff
@@ -25,6 +25,9 @@ func TestConstructFails(t *testing.T) {
 		"/sctp",
 		"/udp/65536",
 		"/tcp/65536",
+		"/onion/9imaq4ygg2iegci7:80",
+		"/onion/aaimaq4ygg2iegci7:80",
+		"/onion/timaq4ygg2iegci7",
```
jbenet: also some with:
- negative ports
- chars outside base32

jbenet:
sorry for horrible delay here, finally catching up with my CRs

protocols.go
david415: ah ok

codec.go
david415: well... perhaps. however i'd argue that an onion address without a port number is useless. an onion address could have several ports listening and each port could correspond to a different service hosted on the same tor instance.

codec.go
david415: ah right...

codec.go
jbenet: yeah sounds right. i'm always in favor of explicitness instead of default magic.

david415:
OK... I've made the requested corrections.

jbenet:
LGTM, thanks @david415

jbenet:
Exciting to see `onion` be part of multiaddr :) :+1:

PR #14, opened 2015-09-17
add in support for udt

whyrusleeping:


whyrusleeping:
after looking through the test failures, i realize we have different ideas of how a udt multiaddr should be represented. thoughts?

whyrusleeping:
my thought was that udt addresses would look like: `/ip4/4.5.6.7/udt/2222`, but i think that you were expecting `/ip4/4.5.6.7/udp/2222/udt` which makes sense too.

jbenet:
```
/ip4/4.5.6.7/udp/2222/udt
```
o/ that one

jbenet:
i had support for UTP already, i think so it shouldn't be too difficult to replicate all the utp parts and s/utp/udt

whyrusleeping:
yeah... i ended up clobbering some things to make it work 'my way'. whoops.

whyrusleeping:
nvm. didnt need to change anything apparently.

jbenet:
yeah i thought i had it there already

PR #16, opened 2016-01-17
Feat/modular

whyrusleeping:
This modularizes multiaddr, and adds a value getter method that i've wanted for forever and a half.

### DIFF
--- a/multiaddr.go
+++ b/multiaddr.go
@@ -117,28 +117,15 @@ func (m *multiaddr) Decapsulate(o Multiaddr) Multiaddr {
 var ErrProtocolNotFound = fmt.Errorf("protocol not found in multiaddr")
 
 func (m *multiaddr) ValueForProtocol(code int) (string, error) {
-	protos := m.Protocols()
-	found := -1
-	index := 2
-
-	for i, p := range protos {
-		if code == p.Code {
+	for _, sub := range Split(m) {
+		p := sub.Protocols()[0]
+		if p.Code == code {
 			if p.Size == 0 {
 				return "", nil
 			}
-			found = i
-			break
-		} else {
-			index += 2
-			if p.Size == 0 {
-				index--
-			}
+			return strings.Split(sub.String(), "/")[2], nil
 		}
 	}
 
-	if found == -1 {
-		return "", ErrProtocolNotFound
-	}
-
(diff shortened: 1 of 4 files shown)


multiaddr_test.go
```diff
@@ -300,3 +300,34 @@ func TestEncapsulate(t *testing.T) {
...
+		t.Fatal(err)
+	}
+
+	if fv != exp {
+		t.Fatalf("expected %q for %d in %d, but got %q instead", exp, p, a, fv)
+	}
+}
+
+func TestGetValue(t *testing.T) {
+	a := newMultiaddr(t, "/ip4/127.0.0.1/utp/tcp/5555/udp/1234/utp/ipfs/QmbHVEEepCi7rn7VL7Exxpd2Ci9NNB6ifvqwhsrbRMgQFP")
```
jbenet: Must test it with edge cases too, like:
```go
"/ip4/0.0.0.0" // only one addr
"/ip4/0.0.0.0/ip4/0.0.0.0/ip4/0.0.0.0" // same sub-addr
"/ip4/0.0.0.0/udp/12345/utp" // ending in a no-value one.
```

multiaddr.go
```diff
@@ -113,3 +113,32 @@ func (m *multiaddr) Decapsulate(o Multiaddr) Multiaddr {
 	}
 	return ma
 }
+
+var ErrProtocolNotFound = fmt.Errorf("protocol not found in multiaddr")
+
+func (m *multiaddr) ValueForProtocol(code int) (string, error) {
```
jbenet: This could use `Split(m)`. Or maybe have a version that yields `[]struct{int, string}` -- maybe: `ProtocolValues(m) -- and this could just be:

```go
func (m *multiaddr) ValueForProtocol(code int) (string, error) {
  for _, part := range m.ProtocolValues() {
    if part.Code == code {
      return part.Value
    }
  }
  return nil, ErrProtocolNotFound
}
```

(This method is calling `m.String()` which is an expensive operation anyway: https://github.com/jbenet/go-multiaddr/blob/master/multiaddr.go#L47-L53 )

jbenet:
couple improvements suggested. but overall looks good

whyrusleeping:
@jbenet addressed your concerns, rfm?

PR #17, opened 2016-01-24
fix interfaces so they work in a nicer go style

whyrusleeping:
This PR makes the go-multiaddr interface match my own views on how interfaces in go should work, namely:

1) Constructors should always return a concrete type, not an interface.
  - this means that no extra casting is required to expose extra capabilities of the underlying type. While its not a big deal for the multiaddr type, take a look at the go-datastore package for why this one matters.

And

2) Interface methods should never contain references to the type in question (or really any other types from within the same package)
  - this ensures that the same interface imported from two differently vendored packages implement eachother

### DIFF
--- a/interface.go
+++ b/interface.go
@@ -14,9 +14,6 @@ Multiaddrs have both a binary and string representation.
 
 */
 type Multiaddr interface {
-	// Equal returns whether two Multiaddrs are exactly equal
-	Equal(Multiaddr) bool
-
 	// Bytes returns the []byte representation of this Multiaddr
 	Bytes() []byte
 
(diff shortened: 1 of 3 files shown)


whyrusleeping:
this isnt finished yet as the interface still references the `Protocols` type. The fix for that is to  move the protocols into their own package. (what? go gets better with increased modularity?)

jbenet:
great, this LGTM. this will be a breaking change. 

great observations @whyrusleeping

whyrusleeping:
@jbenet thoughts on making a go-multiaddr-protocols repo? if we do that, and merge this PR after, we should solve a lot of the 'different import path' issues we've been having.

PR #18, opened 2016-02-05
encode unknown protocol addresses as strings

Gaboose:
I can add new protocols from outside of this library with a line like
```go
ma.AddProtocol(ma.Protocol{42, -1, "dns", ma.CodeToVarint(42)})
```
But it fails when I call `NewMultiaddr("/dns/google.com")`, because it doesn't know how to encode "google.com". What if you encode unknown protocol addresses of variable length as bare strings?

This change makes [this library](https://github.com/Gaboose/go-multiaddr-net) work.

### DIFF
--- a/codec.go
+++ b/codec.go
@@ -211,6 +211,13 @@ func addressStringToBytes(p Protocol, s string) ([]byte, error) {
 		return b, nil
 	}
 
+	// otherwise just encode the bare string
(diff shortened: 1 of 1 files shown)


jbenet:
we'd have to length prefix them, see how i proposed unix recently

PR #23, opened 2016-04-20
Adds gx package.json

noffle:


Kubuxu:
package.json was added independently, sorry.

PR #25, opened 2016-04-29
cleanup panics and make NewFromBytes faster

whyrusleeping:
I started out trying to make NewMultiaddrBytes faster and ended up doing a mild refactor to get rid of *some* of this libraries reliance on panics

### DIFF
--- a/codec.go
+++ b/codec.go
@@ -3,7 +3,6 @@ package multiaddr
 import (
 	"encoding/base32"
 	"encoding/binary"
-	"errors"
 	"fmt"
 	"net"
 	"strconv"
@@ -52,27 +51,42 @@ func stringToBytes(s string) ([]byte, error) {
 	return b, nil
 }
 
-func bytesToString(b []byte) (ret string, err error) {
+func validateBytes(b []byte) (err error) {
 	// panic handler, in case we try accessing bytes incorrectly.
-	defer func() {
-		if e := recover(); e != nil {
-			ret = ""
-			switch e := e.(type) {
-			case error:
-				err = e
-			case string:
-				err = errors.New(e)
-			default:
-				err = fmt.Errorf("%v", e)
-			}
+	for len(b) > 0 {
+		code, n, err := ReadVarintCode(b)
+		b = b[n:]
+		p := ProtocolWithCode(code)
+		if p.Code == 0 {
+			return fmt.Errorf("no protocol with code %d", code)
 		}
-	}()
 
+		if p.Size == 0 {
+			continue
+		}
+
+		size, err := sizeForAddr(p, b)
+		if err != nil {
+			return err
+		}
+
+		if len(b) < size {
+			return fmt.Errorf("invalid value for size")
+		}
+		b = b[size:]
+	}
+
+	return nil
+}
+func bytesToString(b []byte) (ret string, err error) {
 	s := ""
 
 	for len(b) > 0 {
+		code, n, err := ReadVarintCode(b)
+		if err != nil {
+			return "", err
+		}
 
-		code, n := ReadVarintCode(b)
 		b = b[n:]
 		p := ProtocolWithCode(code)
 		if p.Code == 0 {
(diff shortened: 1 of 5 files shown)


multiaddr.go
```diff
@@ -78,7 +81,7 @@ func (m *multiaddr) Protocols() []Protocol {
 		ps = append(ps, p)
 		b = b[n:]
 
-		size = sizeForAddr(p, b)
+		size, err := sizeForAddr(p, b)
```
jbenet: check the error too. i know it's supposed to be caught earlier.

codec.go
```diff
@@ -52,27 +51,43 @@ func stringToBytes(s string) ([]byte, error) {
 	return b, nil
 }
 
-func bytesToString(b []byte) (ret string, err error) {
+func validateBytes(b []byte) (err error) {
 	// panic handler, in case we try accessing bytes incorrectly.
```
jbenet: rm comment

codec.go
```diff
@@ -52,27 +51,43 @@ func stringToBytes(s string) ([]byte, error) {
...
-			case error:
-				err = e
-			case string:
-				err = errors.New(e)
-			default:
-				err = fmt.Errorf("%v", e)
-			}
-		}
-	}()
+	for len(b) > 0 {
```
jbenet: is an empty byte buffer a valid multiaddr? this would say yes. (i dont recall)

codec.go
```diff
@@ -52,27 +51,43 @@ func stringToBytes(s string) ([]byte, error) {
...
+		code, n, err := ReadVarintCode(b)
```
jbenet: check error

codec.go
```diff
@@ -52,27 +51,43 @@ func stringToBytes(s string) ([]byte, error) {
 	return b, nil
 }
 
-func bytesToString(b []byte) (ret string, err error) {
+func validateBytes(b []byte) (err error) {
```
jbenet: this function is not tested on its own. we should add it to all the tests wherever we can. it's critical that the behavior of this one is correct.

multiaddr_test.go
```diff
@@ -342,3 +344,21 @@ func TestGetValue(t *testing.T) {
...
+func TestFuzzBytes(t *testing.T) {
+	rand.Seed(time.Now().UnixNano())
+	// Bump up these numbers if you want to stress this
+	buf := make([]byte, 256)
+	for i := 0; i < 2000; i++ {
+		l := rand.Intn(len(buf))
+		rand.Read(buf[:l])
+
+		// just checking that it doesnt panic
+		ma, err := NewMultiaddrBytes(buf[:l])
```
jbenet: should try to fuzz the string version too.

multiaddr_test.go
```diff
@@ -342,3 +344,21 @@ func TestGetValue(t *testing.T) {
 	assertValueForProto(t, a, P_UDP, "12345")
 	assertValueForProto(t, a, P_UTP, "")
 }
+
+func TestFuzzBytes(t *testing.T) {
```
jbenet: A+

multiaddr.go
```diff
@@ -23,11 +23,11 @@ func NewMultiaddr(s string) (Multiaddr, error) {
 // NewMultiaddrBytes initializes a Multiaddr from a byte representation.
 // It validates it as an input string.
 func NewMultiaddrBytes(b []byte) (Multiaddr, error) {
-	s, err := bytesToString(b)
-	if err != nil {
+	if err := validateBytes(b); err != nil {
```
jbenet: is `recover` expensive? or just `panic`? i'd be pretty worried of running this in production still.

Let's add a `recover` here and if it hits, log an error. if we dont see it happen for a while in production we remove it later. It would be way worse to fail hard because of a panic here than the small slowdown this involves

jbenet:
comments above

codec.go
whyrusleeping: I think it technically is. (it has traditionally been valid, that logic hasnt changed here)

multiaddr.go
whyrusleeping: mmm... annoying that the compiler doesnt complain here

multiaddr.go
whyrusleeping: fair. The defer (not the recover) is a tad slow, but its the panic that really kills things. I'll add a basic recover back in

whyrusleeping:
@jbenet feedback addressed

jbenet:
thanks

Issue #7, opened 2014-09-11
Support DNS 


jbenet:
Multiaddrs like these should work:

```
/dns/ipfs.io
/dns/foo.bar.com/http
```

lgierth:
This format in ambiguous regarding IPv4/IPv6. Other options:

```
/ip4/ipfs.io
/dns6/foo.net
/dnsAAAA/bar.com
/dns/baz.org/ip4
```

All of these would roughly resolve to `/ip4/1.2.3.4` and `/ip6/123::4`. The last one has the most appeal imo.

jbenet:
also, which record? just A? or allow TXT to work too:

```
/dns/baz.org/

vs

/dns/baz.org/a/
/dns/baz.org/txt/
```
(hard to do defaults...)

jbenet:
i like `/dns/baz.org/ip4` though that's a bit tricky. 

```sh
/dns/baz.org/ip    # either 4 or 6
/dns/baz.org/ip4  # ipv4
/dns/baz.org/ip4  # ipv6
```

lgierth:
> maybe:
>
> ```
> /dns/baz.org/ip   # either 4 or 6
> /dns/baz.org/ip4  # ipv4
> /dns/baz.org/ip4  # ipv6
> ```

yeah that's what i was after :)

A full example:
```
/dns/bootstrap1.ipfs.io/ip4/tcp/4001/ipfs/QmaCpDMGvV2BGHeYERUEnRQAwe3N8SzbUtfsmvsqQLuvuJ
```

It'd break with the quasi-rule that a protocol is a 2-tuple, but we can say that you MUST know how many segments follow once you see the first one -- e.g. you see `/dns`, and know there will be two more segments.

jbenet:
@lgierth that sounds good to me

> It'd break with the quasi-rule that a protocol is a 2-tuple, but we can say that you MUST know how many segments follow once you see the first one -- e.g. you see /dns, and know there will be two more segments.

Yep, i think that's fine too (deterministic)

jbenet:
i also havent seen other multiaddr impls, so easy to change for us

jbenet:
@lgierth we should probably support record types here.

```
/dns/bootstrap1.ipfs.io/A/ip4
```
?

lgierth:
Do you mean for SRV/TXT/etc. records? Cause A and AAAA always implicate ip4/ip6.

jbenet:
ah right :+1:

[...]

PR #26, opened 2016-05-09
add support for dns

sivachandran:
The changes allow us to use dns as address. The dns addresses should be
in the form

```
    /dns/<name>/<address-protocol>
```

e.g.

```
    /dns/www.google.com/ip4
    /dns/www.yahoo.com/ip6
```

License: MIT
Signed-off-by: Sivachandran sivachandran.p@gmail.com


sivachandran:
@jbenet @whyrusleeping changes for #7

lgierth:
Hey, we thought about how to do /dns for a long time, and we came up with a /dns,/dns4,/dns6 scheme. Whether it's just /dns with RFC6555-like resolution underneath, or explicit /dns4 and /dns6, remains to be decided, but here's the PR: #46 -- your feedback is very welcome.

I'm very sorry about the lack of responses here.

PR #29, opened 2016-07-31
Fix tor onion string conversion

david415:
added code for onion multiaddr string conversion; added a unit test as well. please review.

### DIFF
--- a/codec.go
+++ b/codec.go
@@ -261,11 +261,9 @@ func addressBytesToString(p Protocol, b []byte) (string, error) {
 		return strconv.Itoa(int(i)), nil
 
 	case P_ONION: // tor onion
-		host_bytes := b[:10]
-		port_bytes := b[10:]
-		host := strings.ToLower(base32.StdEncoding.EncodeToString(host_bytes))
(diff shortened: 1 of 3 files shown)


ianopolous:
See discussion on similar PR: https://github.com/jbenet/go-multiaddr/pull/21

ianopolous:
Don't forget the protocols.csv :-) then we're equivalent.

david415:
please see my comments in
https://github.com/multiformats/go-multiaddr/pull/21

- the onion multiaddr should contain both the onion address and the port number. This makes sense since both peices of addressing info are logically at the same transportation level UNLIKE a IP address and TCP port which are located at different layers of the network stack.

david415:
then merge #21 and we're done with this conversation and we can gradually move on to discussing the more interesting parts of the Tor integration

jbenet:
Thanks very much for this, and for hashing out this stuff. great to hear Tor's port is part of the same. I will merge #21

jbenet:
This has been addressed in https://github.com/multiformats/go-multiaddr/pull/21 -- thank you very much for championing this, @david415 @ianopolous

PR #31, opened 2016-09-12
added unix path protocol

jbenet:
This PR:
- introduces the concept of a "path protocol"
- adds `unix` as a "path protocol"

Path Protocols are terminal protocols, similar to varint length
protocols (such as ipfs), but which in this case want to preserve
string addresses with `/` (slashes) in them. Such multiaddrs look
like this:

    /unix/stdio
    /unix/foo/bar/baz.sock
    /ip4/1.2.3.4/tcp/1234/unix/foo.sock
    /ip4/1.2.3.4/tcp/80/ws/a/b/c

It can also be used for websockets, http, and more.

I'm not yet convinced this is the right way to deal with path protocols
because they force the multiaddr to be terminal (i.e. no more addrs can
be added after). Eg these would be ambiguous, so they're defined to be
part of the path protocol

    /unix/a/b/c/ip4/1.2.3.4/tcp/8080
    # this means a "unix" addr with path "/a/b/c/ip4/1.2.3.4/tcp/8080"
    # NOT: a "unix" addr with path "/a/b/c", then an ip and tcp addrs.

### DIFF
--- a/codec.go
+++ b/codec.go
@@ -8,6 +8,7 @@ import (
 	"net"
 	"strconv"
 	"strings"
+	"errors"
 
 	mh "github.com/jbenet/go-multihash"
 )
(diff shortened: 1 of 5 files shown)


lgierth:
I think it makes sense to be able to address unix sockets. But are there any socket options that we might need to add too?

jbenet:
Good question. not sure.

jbenet:
I'm going to merge this because i'm using it in gtihub.com/multiformats/ma-pipe and it's a pain to deal with all the versions and merges. please consider `/unix` experimental and subject to change.

PR #34, opened 2016-09-15
fix udt<->utp inconsistency in protocols.{go,csv}

keks:
`protocols.go` said
```
	Protocol{P_UTP, 0, "utp", CodeToVarint(P_UTP)},
	Protocol{P_UDT, 0, "udt", CodeToVarint(P_UDT)},
```
and `protocols.csv` said
```
301	0	udt
302	0	utp
```

### DIFF
--- a/protocols.csv
+++ b/protocols.csv
@@ -5,8 +5,8 @@ code	size	name
 33	16	dccp
 41	128	ip6
 132	16	sctp
-301	0	udt
-302	0	utp
+301	0	utp
+302	0	udt
 421	V	ipfs
 480	0	http
 443	0	https


lgierth:
LGTM, fixing it in the csv is better than fixing it in the code, since these codes go into the binary representations of multiaddrs and we currently can't really change the codes there.

keks:
yup, that's what I thought

jbenet:
good call 👍

lgierth:
I just noticed that it should have been the other way round: /udt = 301, /utp = 302 -- that's what both multiaddr/protocols.csv and js-multiaddr have.

PR #35, opened 2016-09-16
Fix bugs parsing onion addresses

cpacia:


cpacia:
gah maybe I should've checked the other PRs first.

PR #36, opened 2016-09-16
add shs/350

keks:
Here is a pull request to add the shs protocol suite designed and first implemented in JS by @dominictarr.
It is used by [go-libp2p-shs](/keks/go-libp2p-shs).

### DIFF
--- a/codec.go
+++ b/codec.go
@@ -9,6 +9,7 @@ import (
 	"strconv"
 	"strings"
 
+	b58 "github.com/jbenet/go-base58"
 	mh "github.com/jbenet/go-multihash"
 )
 
@@ -243,6 +244,16 @@ func addressStringToBytes(p Protocol, s string) ([]byte, error) {
 		size := CodeToVarint(len(m))
 		b := append(size, m...)
 		return b, nil
+
+	case P_SHS: // shs
+		// the address is a varint prefixed multihash string representation
+		a := b58.Decode(s[:])
+		if len(a) == 0 {
+			return nil, fmt.Errorf("failed to parse shs addr: %s", s)
+		}
+		size := CodeToVarint(len(a))
+		b := append(size, a...)
+		return b, nil
 	}
 
 	return []byte{}, fmt.Errorf("failed to parse %s addr: unknown", p.Name)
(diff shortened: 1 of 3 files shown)


jbenet:
Great! Maybe add some tests to https://github.com/multiformats/go-multiaddr/blob/master/multiaddr_test.go to make sure it encodes/decodes as expected

keks:
How could I forget that?

multiaddr_test.go
```diff
@@ -374,7 +382,7 @@ func TestFuzzBytes(t *testing.T) {
 }
 
 func randMaddrString() string {
-	good_corpus := []string{"tcp", "ip", "udp", "ipfs", "0.0.0.0", "127.0.0.1", "12345", "QmbHVEEepCi7rn7VL7Exxpd2Ci9NNB6ifvqwhsrbRMgQFP"}
+	good_corpus := []string{"tcp", "ip", "udp", "ipfs", "shs", "0.0.0.0", "127.0.0.1", "12345", "QmbHVEEepCi7rn7VL7Exxpd2Ci9NNB6ifvqwhsrbRMgQFP", ""}
```
jbenet: empty is fine? this is the good corpus so should be a valid `shs` addr, no?

Weird that tests pass... cc @Kubuxu

jbenet:
Rest LGTM

multiaddr_test.go
keks: No, empty is not fine. Copypaste mistake.

I realized that the shs protocol only uses ed25519 keys, so I changed it from length prefixed var size to `ed25519.PublicKeySize` and added length checks in `codec.go`. The test still passes both with the correct and with empty string, though.

Should I import agl's ed25519 just for using that constant or should I copy it into this package? In case I should import it, how? By vendoring, gx or directly?

jbenet: > I realized that the shs protocol only uses ed25519 keys

Ahh, only ed25519? bummer

> The test still passes both with the correct and with empty string, though.

yeah this is weird, and seems incorrect. @Kubuxu thoughts?

> Should I import agl's ed25519 just for using that constant or should I copy it into this package? In case I should import it, how? By vendoring, gx or directly?

if its just that constant, nah, just copy it in, i think.

Kubuxu: We had similar issue in go-ipfs, even empty, no protocol maddr is a valid maddr.

Should we change it?

lgierth: > I realized that the shs protocol only uses ed25519 keys, so I changed it from length prefixed var size to ed25519.PublicKeySize and added length checks in codec.go. The test still passes both with the correct and with empty string, though.

is there a chance that SHS will support other key algorithms in the future? if so, it should stay variable-length. cc @dominictarr

dominictarr: short answer: no.

shs is designed to be maximally metadata private (as is reasonable). so for this reason, there is no ciphersuite negioation. That also means downgrade attacks are simply not possible.
This also means you can't change shs -- to change shs you have to _change it into another protocol_, and then the new protocol gets a new multiformat name.
The plan is, as you are phasing in a new protocol, you just run it in parallel, say, on a different port.
then remove it when enough of the network has upgraded.

Kubuxu:
I would be for making csv registry of transports and then assigning them and keeping parsing in packages with the transports. Which will be possible with https://github.com/multiformats/go-multiaddr/pull/42

keks:
I didn't get your plan from that one sentence but I also don't get what this has to do with this PR. I mean this was laying around for a month now (idk why it wasn't merged).
I say merge it and maybe do #42 and then make this one use the new way of doing it just like all the others.

keks:
Tbh that seems unrelated.

keks:
Is there a chance of this being merged? Is there something I can do to make it happen? paging @jbenet @whyrusleeping

Kubuxu:
@keks as it already conflicts with codec.go file I will pick this up and add it when #42 gets merged. Does it seems good to you?

Kubuxu:
Or when #42 is merged the parser can be part of the shs transport itself.

lgierth:
The protocol number should be added to protocols.csv in multiformats/multiaddr before merging.

PR #41, opened 2016-10-04
update go-multihash

whyrusleeping:


whyrusleeping:
travis is taking forever, verified tests locally. merging.

PR #42, opened 2016-10-14
add Transocoder functionality

Kubuxu:
Allowing multiaddr extensions to have data with

### DIFF
--- a/transcoders.go
+++ b/transcoders.go
@@ -1,0 +1,175 @@
+package multiaddr
+
+import (
+	"encoding/base32"
+	"encoding/binary"
+	"errors"
+	"fmt"
+	"net"
+	"strconv"
+	"strings"
+
+	mh "github.com/multiformats/go-multihash"
+)
+
+type Transcoder interface {
+	StringToBytes(string) ([]byte, error)
+	BytesToString([]byte) (string, error)
+}
+
+type twrp struct {
+	strtobyte func(string) ([]byte, error)
+	bytetostr func([]byte) (string, error)
+}
+
+func (t twrp) StringToBytes(s string) ([]byte, error) {
+	return t.strtobyte(s)
+}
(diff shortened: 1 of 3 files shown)


lgierth:
What is the goal of this PR, and how does it achieve it? Can't review 370 changed lines with only half a sentence of description ;)

Kubuxu:
166*2 lines of the is just code moved

Rest (50) is Transcoder interface that allows to have data in multiaddr without having your Protocol encoding and decoding inside this repo.

lgierth:
Could you split the refactoring and added functionality into two commits?

Kubuxu:
No it isn't possible as then commit won't build and/or pass tests.

Kubuxu:
//cc @whyrusleeping, @lgierth for CR

transcoders.go
```diff
@@ -0,0 +1,175 @@
...
+
+	mh "github.com/multiformats/go-multihash"
+)
+
+type Transcoder interface {
+	StringToBytes(string) ([]byte, error)
+	BytesToString([]byte) (string, error)
+}
+
+type twrp struct {
```
whyrusleeping: maybe provide a function that constructs one of these given two functions?

Kubuxu:
Updated.

Kubuxu:
@whyrusleeping CR pls.

whyrusleeping:
i wish git diffs were smarter. All youre really doing here is moving code around, but git thinks youre rewriting EVERYTHING.

This LGTM, but would love @lgierth to also take a look.

lgierth:
LGTM :+1: 

It'll also simplify potentially adding multiaddr protocols with more than one value.

whyrusleeping:
@Kubuxu coverage decreased. I'm very disapointed. (though this LGTM and i'm gonna merge it)

PR #46, opened 2017-01-18
Add /p2p, /dns4, /dns6, /wss, /p2p-circuit

lgierth:
See multiformats/multiaddr#34

The new /exp-dns is a test balloon for /dns and will be renamed once we're sure it's right.

Also removes protocols.csv so that there's one less source of confusion.


lgierth:
Updated this as per multiformats/multiaddr#37

The updated scope of this PR is: /p2p, /dns4, /dns6, /wss, /libp2p-circuit-relay

lgierth:
@whyrusleeping that travis osx failure is odd:

```
$ make deps
go get github.com/whyrusleeping/gx
package gx/ipfs/QmYDds3421prZgqKbLpEK7T9Aa2eVdQ7o3YarX1LVLdP2J/go-multihash: unrecognized import path "gx/ipfs/QmYDds3421prZgqKbLpEK7T9Aa2eVdQ7o3YarX1LVLdP2J/go-multihash" (import path does not begin with hostname)
make: *** [gx] Error 1
```

lgierth:
Hold off, this needs more

lgierth:
Updated: includes /p2p, /p2p-circuit, /dns4, /dns6, /ws, /wss

whyrusleeping:
i really don't like us having to modify this package every time we make a change to multiaddrs. This package is pluggable, see how i do it in https://github.com/libp2p/go-ws-transport/blob/master/websocket.go

whyrusleeping:
Also, i think @jbenet wants to keep the protocols.csv file in there, just for the purpose of making this sort of change more deliberate.

Kubuxu:
About removing the protocols.csv, multiple times there was a problem because the unified table was not synced with sub-multiformat table.

whyrusleeping:
@Kubuxu are you in favor of, or against the table here?

Kubuxu:
I think I am against, we might want to improve the table in multicodec to say which multiformat is responsible for that codec but duplicating the table to each repo will be problematic.

lgierth:
> i really don't like us having to modify this package every time we make a change to multiaddrs. This package is pluggable, see how i do it in https://github.com/libp2p/go-ws-transport/blob/master/websocket.go

Yep excellent point, thanks :+1: 

What do you think of breaking it up like this?

- /dns4, /dns6: both go-ipfs and go-multiaddr
  - and lazy-remove the one in go-ipfs with the next go-multiaddr update
- /p2p: go-libp2p
  - also migrate /ipfs there, then lazy-remove the one here in go-multiaddr
- /wss: go-ws-transport
- /p2p-circuit: go-libp2p-circuit (the relay impl)

lgierth:
For /dns4 and /dns6 (and /dnsaddr) I started https://github.com/multiformats/go-multiaddr-dns, which already works and also has a CLI (`madns`). Needs tests, and obviously integration with go-libp2p/go-ipfs.

lgierth:
Closing this as all protocols will be handled in other packages.

PR #48, opened 2017-03-18
add QUIC

marten-seemann:
I added QUIC at Code 81, which is the ASCII code point of "Q".


protocols.go
```diff
@@ -26,6 +26,7 @@ const (
 	P_UDP   = 17
 	P_DCCP  = 33
 	P_IP6   = 41
+	P_QUIC  = 81
```
whyrusleeping: 81 SGTM, please also PR it into the table here: https://github.com/multiformats/multicodec/blob/master/table.csv

whyrusleeping:
LGTM

protocols.go
marten-seemann: Doesn't this belong into https://github.com/multiformats/multiaddr?

Kubuxu: Multicodec unified table, so really both.

marten-seemann: Done, I sent two more PRs.

PR #50, opened 2017-03-18
gx publish 1.2.0

marten-seemann:
for the addition of QUIC


whyrusleeping:
I updated the go multihash dep and published (which includes these changes too). Closing this now, thanks @marten-seemann !

PR #53, opened 2017-07-19
P_CIRCUIT: support p2p-circuit addresses

vyzo:


Kubuxu:
We usually register new protocols externally, example: https://github.com/libp2p/go-ws-transport/blob/master/websocket.go#L39

It might make sense for circuit as it isn't just generic transport.

Stebalien:
@Kubuxu, So no multiprotocol registration?

vyzo:
It is kind of necessary to implement dialing from the libp2p-circuit transport.
See also: https://github.com/libp2p/go-addr-util/pull/4

vyzo:
Also, here: https://github.com/libp2p/go-libp2p-circuit/pull/3 (the user)

vyzo:
Just to clarify, the protocol needs to be registerd before it can be added to the list of transports in go-addr-util, otherwise we have an initialization failure.

vyzo:
@Kubuxu 
Alternative is to modify the code in go-addr-util to allow registration of protocols post initialization, which would not require the change.
I can do that too if you think we should adhere to "external protocol registration" policy, but I think it makes sense to add it to the list as it is intended to be a transport.

Kubuxu:
As you can see, the websocket protocol registers late (it is not in this list) but can be in the go-addr-util list.

vyzo:
Upon further thought and discussion, it's more scalable to register the new protocols dynamically and extend go-addr-util to support dynamic registration of transports.

Closing.

PR #58, opened 2017-10-19
store Multiaddrs as strings and don't use interfaces

Stebalien:
1. This saves one level of indirection, one distinct allocation, and a few bytes.
2. This allows Multiaddrs to be used as map keys (we can use this to make our Peerstore more efficient).

This commit isn't as efficient as it could be. We could improve it slightly by using unsafe casts to temporarily treat strings as bytes. However I'm not sure this will improve performance much and, in general, this change will lead to fewer allocations overall.

TL;DR: We have lots of these, make them small and efficient.


multiaddr.go
```diff
@@ -38,39 +53,37 @@ func NewMultiaddrBytes(b []byte) (a Multiaddr, err error) {
...
+	return []byte(m.bytes)
 }
 
-// String returns the string representation of a Multiaddr
-func (m *multiaddr) String() string {
-	s, err := bytesToString(m.bytes)
+// String returns the string representation of this Multiaddr
+// (may panic if internal state is corrupted)
+func (m Multiaddr) String() string {
+	s, err := bytesToString(m.Bytes())
```
whyrusleeping: if we're going down this path, we might has well have `bstrToString`

Stebalien: We'd just have to call `Bytes()` inside it as we can't extract varints from strings.

Stebalien:
So, the *really* annoying (probably deal breaker) thing here is that `Multiaddr` would no longer have a `nil` value (just an empty `Multiaddr{}` value). This breaks *lots* of code.

whyrusleeping:
I'm not actually sure how this is supposed to save us a lot of memory.

whyrusleeping:
I mean, i get map keys, but we're going to have to do a lot more allocations as a result of this (every call to .String() now allocates a full byte slice, as well as the string its generating)

Stebalien:
We shouldn't have to do that many more allocations:

1. We already copied when calling `Bytes()`.
2. That call to `String()` makes *multiple* allocations (it builds the string). One more won't hurt. Also, unless I'm mistaken, we don't use `String()` in performance critical code. If it does become an issue, we can do some unsafe magic to cast the bstr to a byte slice temporarily.

My main concern here was saving space so we didn't have to store each multiaddr twice (once as a map key, once as a map value). However, after profiling again, I'm pretty sure that's not the main issue. The issue appears to be that maps are wasteful (to avoid collisions) and we shouldn't be storing lots of little maps. I'm working on a patch to store multiaddrs in sorted slices instead of maps.

So yeah, I'm going to close this PR. I'd still like to avoid having a pointer to a pointer but we can do that by defining `type Multiaddr []byte` instead (and we get the added bonus that `nil` still works so we don't break everything).

Stebalien:
Note: we should be able to store records for >10,000 peers in ~6MiB (assuming 5 addresses per peer) so there is definitely room for improvement here before we have to fall back on writing stuff to disk.

PR #59, opened 2017-10-22
Add test for Bytes()

lgierth:
I wanted to have a binary representation example for the Readme, so I wrote a test


Stebalien:
One of the linux CI builds is stick in a queue but this has passed all the other builds so I'm just going to merge it before it falls off my radar.

Stebalien:
Mmm.... Tests, tests are delicious. 🍪

PR #62, opened 2017-12-15
store Multiaddrs as strings and don't use interfaces

Stebalien:
1. This saves one level of indirection, one distinct allocation, and a few bytes.
2. This allows Multiaddrs to be used as map keys (we can use this to make our Peerstore more efficient).
3. Removing the interface allows us to serialize these without copying.

This commit isn't as efficient as it could be. We could improve it slightly by using unsafe casts to temporarily treat strings as bytes. However I'm not sure this will improve performance much and, in general, this change will lead to fewer allocations overall.

### DIFF
--- a/codec.go
+++ b/codec.go
@@ -26,7 +26,7 @@ func stringToMultiaddr(s string) (multiaddr, error) {
 		if p.Code == 0 {
 			return "", fmt.Errorf("no protocol with name %s", sp[0])
 		}
-		b.Write(CodeToVarint(p.Code))
+		b.Write(p.VCode)
 		sp = sp[1:]
 
 		if p.Size == 0 { // no length.
--- a/codec.go
+++ b/codec.go
@@ -6,16 +6,16 @@ import (
 	"strings"
 )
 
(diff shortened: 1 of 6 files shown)


whyrusleeping:
@Stebalien why not keep the interface, and just do:
```go
type multiaddr string
```

That way we get all the same benefits, plus the ability to have nil multiaddrs (making the upgrade path easier)

Stebalien:
Note: I'm pretty sure that'll still require an additional allocation/indirection but it should be fairly cheep.

Kubuxu:
I think Go is smart enough and allocates the type header in the same call as the string.

Kubuxu:
> From a bit of quick testing, it looks like this is correct.

correct as in "it adds allocation"?

Stebalien:
Correct as in it adds a level of indirection. Go may do something sneaky with the allocation but I kind of doubt it. Go *could* do something to avoid an extra object to GC (e.g., have some concept of ownership) but the complexity probably isn't worth it.

PR #63, opened 2017-12-16
make Multiaddr a newtype of []bytes

Stebalien:
This allows us to send multiaddrs on the wire without copying (through the `Bytes()` method). It also avoids one allocation/indirection per multiaddr.

This is an alternative (or stop-gap) to #62 as #62 is invasive. Unlike #62, this PR doesn't break `Multiaddr == nil` or `maddr1.Equals(maddr2)`. However, it *doesn't* allow multiaddrs to be used as map keys :(.

Motivation: Make the allocate less when reporting closer peers. That's currently one of our largest sources of ephemeral byte slice allocations.


multiaddr_test.go
```diff
@@ -215,12 +215,6 @@ func TestBytesSplitAndJoin(t *testing.T) {
 			t.Errorf("joined components failed: %s != %s", m, joined)
 		}
 
-		// modifying underlying bytes is fine.
-		m2 := m.(*multiaddr)
-		for i := range m2.bytes {
```
whyrusleeping: could change this to just directly modify m2. I think the check is still valuable.

Stebalien: No, this is actually broken, that's why I removed it. I'm no longer copying on `Split`, just slicing. I'm now assuming that multiaddrs are immutable. I'm not entirely sure why we had this test in the first place.

PR #66, opened 2018-05-24
begin transition from /ipfs/ to /p2p/ multiaddrs

whyrusleeping:
This is a little hacky, and needs tests, but this is the rightish way forward

### DIFF
--- a/protocols.go
+++ b/protocols.go
@@ -33,7 +33,8 @@ const (
 	P_UDT   = 0x012D
 	P_UTP   = 0x012E
 	P_UNIX  = 0x0190
-	P_IPFS  = 0x01A5
+	P_P2P   = 0x01A5
+	P_IPFS  = 0x01A5 // alias for backwards compatability
 	P_HTTP  = 0x01E0
 	P_HTTPS = 0x01BB
 	P_ONION = 0x01BC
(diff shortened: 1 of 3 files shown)


cpacia:
Is /p2p supposed to be only used for addressing or also representing IPFS file locations?

I guess along those same lines is /ipfs going to be completely deprecated? I'm currently using a /ipfs multiaddr in my code to represent file location. I assume I'm probably the only one using it that way.

protocols.go
```diff
@@ -45,21 +46,76 @@ const (
...
+// The first stage of the rollout is to ship this package to all users so
+// that all users of multiaddr can parse both /ipfs/ and /p2p/ multiaddrs
+// as the same code (P_P2P). During this stage of the rollout, all addresses
+// with P_P2P will continue printing as /ipfs/, so that older clients without
+// the new parsing code won't break.
+// Once the network has adopted the new parsing code broadly enough, users of
+// multiaddr can add a call to this method to an init function in their codebase.
+// This will cause any P_P2P multiaddr to print out as /p2p/ instead of /ipfs/.
+// Note that the binary serialization of this multiaddr does not change at any
+// point. This means that this code is not a breaking network change at any point
```
VictorBjelkholm: I think I've missed a memo about this. We should make sure we're doing these stages of rollout at the same time in JS-land, correct?

whyrusleeping:
@cpacia so the goal here is to deprecate multiaddrs that contain /ipfs/, as a multiaddr is a network endpoint description, and what we really mean when we have an /ipfs/ multiaddr is 'this is a libp2p node'. Quite a few projects other than ipfs are using libp2p node, so having them all have to use /ipfs/ all over is a bit odd.

Multiaddrs can't actually represent the same information as the /ipfs/ file namespace, as the multiaddr only supports a bare multihash, and the ipfs file space supports CIDs (which contain bare multihashes as a valid value, for backwards compatibility).

Where are you using a multiaddr to reference a file? Let's figure out how to move forward with that and transition is to use a CID instead

protocols.go
Stebalien: This stage can go forward at any time (the sooner the better). However, we'll need to wait for JS-land before moving on to the second stage.

protocols.go
```diff
@@ -73,17 +129,13 @@ func AddProtocol(p Protocol) error {
...
 
 // ProtocolWithName returns the Protocol description with given string name.
 func ProtocolWithName(s string) Protocol {
-	for _, p := range Protocols {
-		if p.Name == s {
-			return p
-		}
-	}
-	return Protocol{}
+	return ProtocolsByName[s]
```
Stebalien: :heart:

Stebalien:
:rocket:

Stebalien:
(tested against go-ipfs)

PR #69, opened 2018-06-16
Add "ip6%" multiaddr (IPv6 with zone ID)

mwnx:
---

See: https://github.com/multiformats/multiaddr/pull/68

### DIFF
--- a/multiaddr_test.go
+++ b/multiaddr_test.go
@@ -23,6 +23,9 @@ func TestConstructFails(t *testing.T) {
 		"/ip4/::1",
 		"/ip4/fdpsofodsajfdoisa",
 		"/ip6",
+		"/ip6zone",
+		"/ip6zone/",
+		"/ip6zone//ip6/fe80::1",
 		"/udp",
 		"/tcp",
 		"/sctp",
@@ -65,6 +68,10 @@ func TestConstructSucceeds(t *testing.T) {
 		"/ip6/::1",
 		"/ip6/2601:9:4f81:9700:803e:ca65:66e8:c21",
 		"/ip6/2601:9:4f81:9700:803e:ca65:66e8:c21/udp/1234/quic",
+		"/ip6zone/x/ip6/fe80::1",
+		"/ip6zone/x%y/ip6/fe80::1",
+		"/ip6zone/x%y/ip6/::",
+		"/ip6zone/x/ip6/fe80::1/udp/1234/quic",
 		"/onion/timaq4ygg2iegci7:1234",
 		"/onion/timaq4ygg2iegci7:80/http",
 		"/udp/0",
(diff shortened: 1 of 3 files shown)


Stebalien:
So, I added some validation logic to the zone but I'm not sure if we *can* do that. The zone can technically include anything. How do you feel about URL escaping (in the string format)?

mwnx:
The standard does not explicitly ban the use the `/` character in the zone ID. However, it does allow combining zone IDs with the `/<subnet>` syntax and gives the following example:
```
            fe80::%2/64
```
This would imply that allowing slashes could lead to ambiguity in situations where a program accepts either an address or a prefix.

So, until we specify a way of properly handling `/` characters in multiaddr values, I'd say the validation code that you added is an acceptable compromise. Are there any plans for dealing with that issue by the way?

Thanks for fixing my PR by the way. :)

Stebalien:
> Are there any plans for dealing with that issue by the way?

We handle this on a per-protocol basis.

mwnx:
On Mon, Sep 17, 2018 at 02:13:58PM -0700, Steven Allen wrote:
> > Are there any plans for dealing with that issue by the way?
> 
> We handle this on a per-protocol basis.

I can't seem to find any concrete example of slash handling in the
current code. What do you propose in this case? Taking the PR as-is
(for now at least) would be fine with me. Escaping of slashes could
be added later. But if we do add escaping, what escaping scheme
should we use? Has this been discussed at all for other protocols
(e.g. unix, http)?

-- 
mwnx
GPG: AEC9 554B 07BD F60D 75A3  AF6A 44E8 E4D4 0312 C726

Stebalien:
> I can't seem to find any concrete example of slash handling in the current code.

Checkout the go-multiaddr-dns package.

> Escaping of slashes could be added later. But if we do add escaping, what escaping scheme should we use? Has this been discussed at all for other protocols (e.g. unix, http)?

This has not been discussed. Note: `unix` is actually a terminal *path* protocol. That is, everything after `/unix/...` is considered to be an argument to `/unix` (we really should reconsider that before we start using it as it limits it's usefulness). `http` doesn't *take* an argument (at the moment). It just means "talk to this HTTP server".

For things that usually *aren't* paths (e.g., zones), I'd just make escaping part of the spec. Unfortunately, that doesn't work for protocols like `/unix/`. Personally, I'd kind of like to introduce a bracket syntax (`/unix/[/a/b/c/d/]/ipfs/QmId`) but that syntax is a bit hard to visually parse.

If you have any thoughts or ideas, please open an issue to discuss it. Unfortunately, many of us don't have time to think through some of these design questions enough to actually come up with reasonable solutions.

Stebalien:
Merged. As you pointed out, there's no point in holding this patch back on this discussion.

PR #71, opened 2018-06-22
fully validate multiaddr byte representations on construction

Stebalien:
Given that we have to modify transcoders anyways, this patch also moves some
repetitive logic out of these transcoders into this library.

### DIFF
--- a/codec.go
+++ b/codec.go
@@ -26,7 +26,7 @@ func stringToBytes(s string) ([]byte, error) {
 		if p.Code == 0 {
 			return nil, fmt.Errorf("no protocol with name %s", sp[0])
 		}
-		b.Write(CodeToVarint(p.Code))
+		_, _ = b.Write(CodeToVarint(p.Code))
 		sp = sp[1:]
 
 		if p.Size == 0 { // no length.
@@ -51,7 +51,7 @@ func stringToBytes(s string) ([]byte, error) {
 			return nil, fmt.Errorf("failed to parse %s: %s %s", p.Name, sp[0], err)
 		}
 		if p.Size < 0 { // varint size.
-			b.Write(CodeToVarint(len(a)))
+			_, _ = b.Write(CodeToVarint(len(a)))
 		}
 		b.Write(a)
 		sp = sp[1:]
--- a/codec.go
+++ b/codec.go
@@ -50,6 +50,9 @@ func stringToBytes(s string) ([]byte, error) {
 		if err != nil {
 			return nil, fmt.Errorf("failed to parse %s: %s %s", p.Name, sp[0], err)
 		}
+		if p.Size < 0 { // varint size.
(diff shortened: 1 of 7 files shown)


codec.go
```diff
@@ -50,6 +50,9 @@ func stringToBytes(s string) ([]byte, error) {
 		if err != nil {
 			return nil, fmt.Errorf("failed to parse %s: %s %s", p.Name, sp[0], err)
 		}
+		if p.Size < 0 { // varint size.
+			b.Write(CodeToVarint(len(a)))
```
whyrusleeping: maybe explicitly throw away the error values?

transcoders.go
```diff
@@ -121,61 +131,36 @@ func onionBtS(b []byte) (string, error) {
...
-	if len(b) != size {
-		return "", errors.New("inconsistent lengths")
-	}
-	if size == 0 {
-		return "", errors.New("invalid length")
-	}
-	s := string(b)
-	s = s[1:] // remove starting slash
-	return s, nil
+	return string(b), nil
```
whyrusleeping: reasoning here?

transcoders.go
```diff
@@ -121,61 +131,36 @@ func onionBtS(b []byte) (string, error) {
...
 
-var TranscoderUnix = NewTranscoderFromFunctions(unixStB, unixBtS)
+var TranscoderUnix = NewTranscoderFromFunctions(unixStB, unixBtS, nil)
 
 func unixStB(s string) ([]byte, error) {
-	// the address is the whole remaining string, prefixed by a varint len
-	size := CodeToVarint(len(s))
-	b := append(size, []byte(s)...)
-	return b, nil
+	return []byte(s), nil
```
whyrusleeping: are we dropping the varints for some reason? I assume youre not changing the binary format, so where is this being moved to?

transcoders.go
Stebalien: I wanted it to be symmetric (that is, the bytes to string function should be an inverse of the string to bytes function). I'm now handling this in `bytesToString` (as we do in `stringToBytes`).

transcoders.go
Stebalien: We're now doing this in stringToBytes and bytesToString. Before, we were doing this for *every* protocol with a variable length size (and we were doing the exact same thing. On the other hand, we had *other* code that assumed that we were using varint length prefixes so I figured we might as well just centralize this logic and simplify things.

This change and the path one are both to pave the way for nice multiaddr constructors/decoders (so we can stop parsing strings).

transcoders.go
Stebalien: We could also not prepend a slash here:

https://github.com/multiformats/go-multiaddr/blob/84fb105635d33c95b41a9385064329b663f6f56f/codec.go#L43

Issue #64, opened 2018-03-17
Update the multicodecs list according to latest table


daviddias:
https://github.com/multiformats/multiaddr/blob/master/protocols.csv#L22-L25

Needed for https://github.com/ipfs/interop/pull/6

//cc @Stebalien

PR #72, opened 2018-07-13
feat: Adds ws and wss multiaddr

gpestana:
Adds `ws` and `wss` multiaddr as defined in https://github.com/multiformats/multiaddr/blob/master/protocols.csv

Related to #64

### DIFF
--- a/multiaddr_test.go
+++ b/multiaddr_test.go
@@ -98,6 +98,10 @@ func TestConstructSucceeds(t *testing.T) {
 		"/ip4/1.2.3.4/tcp/80/unix/a/b/c/d/e/f",
 		"/ip4/127.0.0.1/ipfs/QmcgpsyWgH8Y8ajJz1Cu72KnS5uo2Aa2LpzU7kinSupNKC/tcp/1234/unix/stdio",
 		"/ip4/127.0.0.1/p2p/QmcgpsyWgH8Y8ajJz1Cu72KnS5uo2Aa2LpzU7kinSupNKC/tcp/1234/unix/stdio",
+		"/ip4/127.0.0.1/tcp/127/ws",
+		"/ip4/127.0.0.1/tcp/127/ws/",
+		"/ip4/127.0.0.1/tcp/127/wss",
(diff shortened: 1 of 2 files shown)


Stebalien:
The WS protocol is currently defined in go-ws-transport (but we don't define the wss protocol because we can't use it).

However, I understand the issue.

@whyrusleeping

Issue: We can neither parse nor print a multiaddr unless we understand the full multiaddr.
Solutions:

1. This PR (put every protocol definition we can in one repo).
2. Add an "unknown" protocol to the string format: https://github.com/multiformats/multiaddr/issues/70

The second completely solves the issue that JS is having but the first one is (a) easier and (b) provides better usability.

gpestana:
The 2) option sounds reasonable to me, although I might not fully understand what are the issues JS is having with the option 1). I think that if the WS protocol is implemented somewhere else, it should be documented somewhere here as it gets confusing for newcomers. 

What would be the criteria to add a new protocol to the list and implement it vs treat it as `unknown`?

Stebalien:
The current issue for javascript land is that go-multiaddr doesn't understand protocols like `wss` or `webrtc` (?). This means that:

1. We *throw away* these addresses when we encounter them (in the DHT).
2. We can't dial addresses like `/ip4/4.3.2.1/tcp/321/p2p/QmRelay/p2p-circuit/ip4/1.2.3.4/tcp/123/wss/ipfs/QmDest` *even if* we speak all the protocols *we* need to speak (those before `p2p-circuit`).

Adding these protocols to this package will fix these issues for now. However, the second we add a new protocol, we'll have the same problem.

We can fix 1 by simply not decoding these addresses. However, that would be a large, weird change and I'd like to avoid it if possible. Note: that issue will be fixed eventually once we get https://github.com/ipfs/notes/issues/291#issuecomment-396003860 but that's a ways off.

Fix 2 (the "unknown" protocol) fixes issue 2 from above. That's basically the only fix I can think of.

Stebalien:
We went with option 1 (define everything here) for now. Updated PR in #122.

PR #73, opened 2018-07-18
add protocol by code map

Stebalien:
and unexport the protocol by name map.

This also checks protocol definitions on *add* instead of at runtime.

### DIFF
--- a/codec.go
+++ b/codec.go
@@ -155,12 +155,6 @@ func sizeForAddr(p Protocol, b []byte) (skip, size int, err error) {
 		return 0, (p.Size / 8), nil
 	case p.Size == 0:
 		return 0, 0, nil
-	case p.Path:
-		size, n, err := ReadVarintCode(b)
-		if err != nil {
-			return 0, 0, err
-		}
-		return n, size, nil
 	default:
 		size, n, err := ReadVarintCode(b)
 		if err != nil {
--- a/protocol.go
+++ b/protocol.go
@@ -1,0 +1,129 @@
+package multiaddr
+
+import (
+	"fmt"
+	"strings"
+)
+
+// These are special sizes
+const (
+	LengthPrefixedVarSize = -1
+)
+
+// Protocol is a Multiaddr protocol description structure.
+type Protocol struct {
+	// Name is the string representation of the protocol code. E.g., ip4,
+	// ip6, tcp, udp, etc.
+	Name string
+
+	// Code is the protocol's multicodec (a normal, non-varint number).
+	Code int
+
+	// VCode is a precomputed varint encoded version of Code.
+	VCode []byte
+
+	// Size is the size of the argument to this protocol.
+	//
+	// * Size == 0 means this protocol takes no argument.
+	// * Size >  0 means this protocol takes a constant sized argument.
+	// * Size <  0 means this protocol takes a variable length, varint
+	//             prefixed argument.
+	Size int // a size of -1 indicates a length-prefixed variable size
+
+	// Path indicates a path protocol (e.g., unix). When parsing multiaddr
+	// strings, path protocols consume the remainder of the address instead
+	// of stopping at the next forward slash.
+	//
+	// Size must be LengthPrefixedVarSize.
+	Path bool
+
+	// Transcoder converts between the byte representation and the string
+	// representation of this protocol's argument (if any).
+	//
+	// This should only be non-nil if Size != 0
+	Transcoder Transcoder
+}
+
+var protocolsByName = map[string]Protocol{}
+var protocolsByCode = map[int]Protocol{}
+
+// Protocols is the list of multiaddr protocols supported by this module.
+var Protocols = []Protocol{}
+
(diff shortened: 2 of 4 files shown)


codec.go
```diff
@@ -155,12 +155,6 @@ func sizeForAddr(p Protocol, b []byte) (skip, size int, err error) {
 		return 0, (p.Size / 8), nil
 	case p.Size == 0:
 		return 0, 0, nil
-	case p.Path:
```
Stebalien: Redundant check.

protocol.go
```diff
@@ -0,0 +1,129 @@
...
+	"fmt"
+	"strings"
+)
+
+// These are special sizes
+const (
+	LengthPrefixedVarSize = -1
+)
+
+// Protocol is a Multiaddr protocol description structure.
```
Stebalien: Moved and documented.

protocol.go
```diff
@@ -0,0 +1,129 @@
...
+func AddProtocol(p Protocol) error {
+	if _, ok := protocolsByName[p.Name]; ok {
+		return fmt.Errorf("protocol by the name %q already exists", p.Name)
+	}
+
+	if _, ok := protocolsByCode[p.Code]; ok {
+		return fmt.Errorf("protocol code %d already taken by %q", p.Code, p.Code)
+	}
+
+	if p.Size != 0 && p.Transcoder == nil {
```
Stebalien: Check everything up-front.

protocol.go
```diff
@@ -0,0 +1,129 @@
...
+}
+
+func AddProtocol(p Protocol) error {
+	if _, ok := protocolsByName[p.Name]; ok {
+		return fmt.Errorf("protocol by the name %q already exists", p.Name)
+	}
+
+	if _, ok := protocolsByCode[p.Code]; ok {
+		return fmt.Errorf("protocol code %d already taken by %q", p.Code, p.Code)
+	}
```
Stebalien: O(1) duplicate checks...

protocols.go
```diff
@@ -40,156 +23,126 @@ const (
...
-	protoUNIX,
-}
-
 var (
-	protoIP4  = Protocol{P_IP4, 32, "ip4", CodeToVarint(P_IP4), false, TranscoderIP4}
-	protoTCP  = Protocol{P_TCP, 16, "tcp", CodeToVarint(P_TCP), false, TranscoderPort}
-	protoUDP  = Protocol{P_UDP, 16, "udp", CodeToVarint(P_UDP), false, TranscoderPort}
-	protoDCCP = Protocol{P_DCCP, 16, "dccp", CodeToVarint(P_DCCP), false, TranscoderPort}
-	protoIP6  = Protocol{P_IP6, 128, "ip6", CodeToVarint(P_IP6), false, TranscoderIP6}
+	protoIP4 = Protocol{
```
Stebalien: Easier to read.

protocols.go
```diff
@@ -40,156 +23,126 @@ const (
...
+	} {
+		if err := AddProtocol(p); err != nil {
+			panic(err)
+		}
 	}
-	return int(num), n, nil
+
+	// explicitly set both of these
+	protocolsByName["p2p"] = protoP2P
+	protocolsByName["ipfs"] = protoP2P
```
gpestana: `protocolsByCode[protoP2P.Code] = protoP2P` missing?

protocols.go
gpestana: Yes, much more readable now 👍

protocols.go
Stebalien: We set that in `AddProtocol`. Really, the "ipfs" one here isn't necessary. We only need this to set the "p2p" alias.

Stebalien:
I'll count that as a review.

cretz:
This change broke the ability for external packages to override a protocol by name (the map was exposed previously, now it is hidden). Is this by intention to not support protocol overrides by external packages?

Stebalien:
> This change broke the ability for external packages to override a protocol by name (the map was exposed previously, now it is hidden). Is this by intention to not support protocol overrides by external packages?

What do you mean by "override by name"? Protocols with the same name should have the same definition. If the current protocol definition doesn't fit your needs, we should define a new one. Redefining an existing protocol will just introduce incompatibilities.

cretz:
@Stebalien - Yeah, as a hack I was overridding the onion one because it is no good for v3 addresses (ref https://github.com/multiformats/multiaddr/issues/65). It was in my own bootstrapped DHT so I was not concerned w/ compatibility, but in retrospect, I should just probably add a new protocol and just stop using `/onion`.

cretz:
@Stebalien - I remember why I did it now, it's because the JS side doesn't easily allow you do add one dynamically (ref https://github.com/multiformats/js-multiaddr/issues/22). So between the not-all-onion-supporting onion protocol and the JS lib's lack of extensibility, there is very little that those that want to use libp2p across languages can do here unless they step in and fix/fork these things.

Stebalien:
If you have time, I'd submit a PR to both go and js for a new `onion3` protocol. I can push on the JS side (if you bug me) to get it merged.

PR #74, opened 2018-07-18
[WIP] add the "unknown" multiaddr

Stebalien:
This should be considered carefully. We may want to modify the API to support it.

implements https://github.com/multiformats/go-multiaddr/pull/73

### DIFF
--- a/protocol.go
+++ b/protocol.go
@@ -17,6 +17,8 @@ type Protocol struct {
 	Name string
 
 	// Code is the protocol's multicodec (a normal, non-varint number).
+	//
+	// -1 is reserved for the special "unknown" protocol.
 	Code int
 
 	// VCode is a precomputed varint encoded version of Code.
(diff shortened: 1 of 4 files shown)


protocol.go
```diff
@@ -17,6 +17,8 @@ type Protocol struct {
 	Name string
 
 	// Code is the protocol's multicodec (a normal, non-varint number).
+	//
+	// -1 is reserved for the special "unknown" protocol.
```
gpestana: This comment seems to be inconsistent with the rest of the code. e.g.:

```
ps = append(ps, Protocol{
  Name: "unknown",
  Size: -1,
  Code: 0,
})
```

IMO, -1 for unknown protocol fells better.

Stebalien: Sorry, I changed my mind multiple times half-way through. I agree.

PR #80, opened 2018-10-01
add a Component interface to multiaddrs

Stebalien:
This adds a `Component` interface to multiaddrs and deprecates `ValueForProtocol`. It should make working with multiaddrs significantly less painful.

Eventually, I'd like to add different component and multiaddr types (like the `net.Addr` types). However, I'm going to do this step-by-step.

### DIFF
--- a/codec.go
+++ b/codec.go
@@ -96,51 +96,55 @@ func validateBytes(b []byte) (err error) {
 	return nil
 }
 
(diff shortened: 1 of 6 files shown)


Stebalien:
This doesn't have the desired performance improvements.

Stebalien:
This doesn't have the desired performance improvements.

PR #81, opened 2018-10-02
add component/foreach helpers

Stebalien:
This adds a `Component` helper type and a `ForEach` helper method.

The first attempt used an interface but interfaces imply allocation. We really
can't afford to allocate here.

### DIFF
--- a/component.go
+++ b/component.go
@@ -130,20 +130,3 @@ func newComponent(protocol Protocol, bvalue []byte) *Component {
 		offset:   offset,
 	}
 }
-
-// ForEach walks over the multiaddr, component by component.
-//
-// This function iterates over components *by value* to avoid allocating.
-func ForEach(m Multiaddr, cb func(c Component) bool) {
-	b := m.Bytes()
-	for len(b) > 0 {
-		n, c, err := readComponent(b)
-		if err != nil {
-			panic(err)
-		}
-		if !cb(c) {
-			return
-		}
-		b = b[n:]
-	}
-}
--- a/component.go
+++ b/component.go
@@ -1,0 +1,149 @@
+package multiaddr
+
+import (
+	"bytes"
+	"encoding/binary"
+	"fmt"
+	"strings"
+)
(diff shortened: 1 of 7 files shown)


Stebalien:
Note: this changes the behavior of paths. `ValueForProtocol(pathProtocol)` will now make the path start with a `/`.

Rational:

1. Paths should start with `/`.
2. This makes the string-to-bytes and bytes-to-string encoders symmetric. In the past, the former used to take a path starting with a `/` and the latter would spit out a path without the `/`.

component.go
```diff
@@ -0,0 +1,149 @@
...
+
+func (c *Component) Equal(o Multiaddr) bool {
+	return bytes.Equal(c.bytes, o.Bytes())
+}
+
+func (c *Component) Protocols() []Protocol {
+	return []Protocol{c.protocol}
+}
+
+func (c *Component) Decapsulate(o Multiaddr) Multiaddr {
```
raulk: I guess `Decapsulate` doesn't make much sense in the context of a single component, so it becomes equivalent to a Filter operation, i.e. returns itself if there's a match?

component.go
```diff
@@ -0,0 +1,149 @@
...
+		bytes:    maddr,
+		protocol: protocol,
+		offset:   offset,
+	}
+}
+
+// ForEach walks over the multiaddr, component by component.
+//
+// This function iterates over components *by value* to avoid allocating.
+func ForEach(m Multiaddr, cb func(c Component) bool) {
```
raulk: Would it make sense to add a `Filter` function that returns the components for a given protocol?

Stebalien: Not sure about filter but I'm adding `SplitFirst`, `SplitLast`, `SplitFunc`, etc. functions.

component.go
Stebalien: Basically. Personally, given the new `SplitFunc`, I'm not sure how useful decapsulate is in general.

PR #82, opened 2018-10-02
Add HasPrefix() utility function

lgierth:
I sketched this out last night, it seems useful on first sight, but I don't have the time to work it out properly. Consider it up for grabs :)

### DIFF
--- a/util.go
+++ b/util.go
@@ -54,3 +54,23 @@ func StringCast(s string) Multiaddr {
 	}
 	return m
 }
+
+func HasPrefix(addr, prefix Multiaddr) bool {
+	protos := addr.Protocols()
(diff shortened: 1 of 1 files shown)


Stebalien:
Does `bytes.IsPrefix(a.Bytes(), b.Bytes())` not work?

lgierth:
This one here works on the protocol and value boundaries. But hrm yeah I suppose if we're dealing with a multiaddr created through the proper functions, `Bytes()` will be on the boundaries too :)

Stebalien:
(although it might be reasonable to add a function/method)

Stebalien:
(but I'd rather not leave a PR around for that, we can just put it in an issue)

PR #87, opened 2018-11-29
onionv3 and i2p base64 addresses

eyedeekay:
I seem to have successfully added onionv3 addresses and i2p base64 addresses. I elected to use the base64 addresses for i2p because I can use it to load the keys in a form which is usable by https://github.com/kpetku/sam3, which one of the existing Go libraries for interacting with i2p. There are several options which can compute the base32 key from the base64 key, including sam3 itself, but if need be I could also implement i2p base32 keys as their own thing.


transcoders.go
```diff
@@ -167,6 +169,94 @@ func onionBtS(b []byte) (string, error) {
...
+func garlicStB(s string) ([]byte, error) {
+	addr := strings.Split(s, ":")
+	if len(addr) != 2 {
+		return nil, fmt.Errorf("failed to parse garlic addr: %s does not contain a port number.", s)
+	}
+	// i2p base64 address
+	if len(addr[0]) != 516 {
+		return nil, fmt.Errorf("failed to parse garlic addr: %s not an i2p base64 address. len: %d\n", s, len(addr[0]))
+	}
+	garlicHostBytes, err := base64.NewEncoding("ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789-~").DecodeString(addr[0])
```
Stebalien: We should define this outside.

transcoders.go
```diff
@@ -167,6 +169,94 @@ func onionBtS(b []byte) (string, error) {
...
+	binary.BigEndian.PutUint16(garlicPortBytes, uint16(i))
+	bytes := []byte{}
+	bytes = append(bytes, garlicHostBytes...)
+	bytes = append(bytes, garlicPortBytes...)
+
+	return bytes, nil
+}
+
+func garlicBtS(b []byte) (string, error) {
+	addr := base64.NewEncoding("ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789-~").EncodeToString(b[0:387])
```
Stebalien: ditto.

transcoders.go
```diff
@@ -167,6 +168,93 @@ func onionBtS(b []byte) (string, error) {
...
+	}
+
+	bytes := []byte{}
+	bytes = append(bytes, garlicHostBytes...)
+
+	return bytes, nil
+}
+
+func garlic64BtS(b []byte) (string, error) {
+	addr := garlicBase64Encoding().EncodeToString(b)
```
Stebalien: For completeness, we should check the length.

transcoders.go
```diff
@@ -167,6 +168,93 @@ func onionBtS(b []byte) (string, error) {
...
+
+func garlicValidate(b []byte) error {
+	if len(b) < 386 {
+		return fmt.Errorf("failed to validate garlic addr: %s not an i2p base64 address. len: %d\n", b, len(b))
+	}
+	s, err := garlic64BtS(b)
+	if err != nil {
+		return fmt.Errorf("failed to decode base64 i2p addr: %s %s", s, err)
+	}
+	_, err = garlicBase64Encoding().DecodeString(s)
```
Stebalien: Do we need to encode/decode? Validate's supposed to be a fast subset of `garlic64BtS` (one that only *checks* if the bytes are correct without actually turning them into a string). I *believe* simply checking the length would be sufficient.

transcoders.go
```diff
@@ -167,6 +168,93 @@ func onionBtS(b []byte) (string, error) {
...
+	if len(s) < 516 || len(s) > 616 {
+		return nil, fmt.Errorf("failed to parse garlic addr: %s not an i2p base64 address. len: %d\n", s, len(s))
+	}
+	garlicHostBytes, err := garlicBase64Encoding().DecodeString(s)
+	if err != nil {
+		return nil, fmt.Errorf("failed to decode base64 i2p addr: %s %s", s, err)
+	}
+
+	bytes := []byte{}
+	bytes = append(bytes, garlicHostBytes...)
```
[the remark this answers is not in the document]
eyedeekay: Nope. Relic of the old approach I missed.

Stebalien:
Awesome! Thanks for sticking with me through this process!

PR #88, opened 2018-11-29
Fix bug in VarintSize

bigs:
`VarintSize` has a bug that makes it return the wrong result for all unsigned varints requiring a multiple of 7 bits to represent. This fixes that. So far as I know, this is only used by `NewComponent`, but since it is publicly exported I figured it was worth patching rather than removing it outright.

PS: Could I be added to the multiformats org/go team?


Stebalien:
Nice catch!

1. Is the function immediately below this incorrect as well? That's where this logic came from. (or does this over estimate).
2. Mind adding a test?

bigs:
ahh, yep! it should, in fact, rely on it. it will overestimate and return a buffer larger than it needs to be. i'll fix that and add a test. this was one of those days where i went in to add unix socket support for `manet` and wound up here...

bigs:
@Stebalien naturally in writing tests i found another bug... 😭. sorted.

varint_test.go
```diff
@@ -0,0 +1,16 @@
...
+	size := VarintSize(x)
+	if size != expected {
+		t.Fatalf("expected varintsize of %d to be %d, got %d", x, expected, size)
+	}
+}
+
+func TestVarintSize(t *testing.T) {
+	expectVarint(t, (1<<7)-1, 1)
+	expectVarint(t, 0, 1)
+	expectVarint(t, 1<<7, 2)
```
Stebalien: What about just computing all varints 0-1024? (or something like that).

PR #92, opened 2018-12-20
Add dns and p2p-webrtc-direct protocols

backkem:
This PR ports the `dns` and `p2p-webrtc-direct` protocol from `js-multiaddr`. This is in preparation of adding the `WebRTCDirect` pattern to `whyrusleeping/mafmt`.


backkem:
Will investigate overlap with `multiformats/go-multiaddr-dns` and reopen.

PR #95, opened 2019-02-25
Let Multiaddr implement marshalers: binary, text, json

hsanjuan:
I don't think there is a way to do it without making `multiaddr` a pointer for all methods, since we need to satisfy the interface.


hsanjuan:
Thank you for the quick review @Stebalien !

PR #101, opened 2019-04-05
I2P Base32 and Base32 for Encrypted Leasesets multiaddr, tiny garlic64 fix

eyedeekay:
This makes go-multiaddr compatible with i2p base32 addresses and also corrects the garlic64 addresses, somehow a disagreement between the protocol code in use by the library and the protocol code listed by multiaddr slipped by me when I initially created it. 

Here's a link to the proposal for base32 for encrypted leasesets, it reflects a significant change to how base32's work in some cases and includes changing them to be variable-length. It mostly doesn't apply here but since up to very recently all base32's were of a fixed length it seemed potentially pertinent.

http://www.i2p.net/spec/proposals/149-b32-encrypted-ls2


transcoders.go
```diff
@@ -238,13 +238,66 @@ func garlic64BtS(b []byte) (string, error) {
...
 	}
 	return nil
 }
 
+var TranscoderGarlic32 = NewTranscoderFromFunctions(garlic32StB, garlic32BtS, garlic32Validate)
+
+var garlicBase32Encoding = base32.NewEncoding("abcdefghijklmnopqrstuvwxyz234567")
+
+func garlic32StB(s string) ([]byte, error) {
+	s = strings.Replace(s, ".b32.i2p", "", -1)
```
Stebalien: `strings.TrimSuffix`

But really, do we want to allow this? Personally, I wouldn't expect this suffix to be included in the multiaddr.

transcoders.go
```diff
@@ -238,13 +238,66 @@ func garlic64BtS(b []byte) (string, error) {
...
+	// garlic address without the ".b32.i2p" substring
+	if len(s) < 55 {
+		if len(s) != 52 {
```
Stebalien: Can we comment on what these constants mean and why we're doing this?

transcoders.go
```diff
@@ -238,13 +238,66 @@ func garlic64BtS(b []byte) (string, error) {
...
+			return nil, fmt.Errorf("failed to parse garlic addr: %s not a i2p base32 address. len: %d", s, len(s))
+		}
+	}
+
+	padout := func(s2 string) string {
```
Stebalien: Defining this as a function seems unnecessary.

transcoders.go
```diff
@@ -238,13 +238,66 @@ func garlic64BtS(b []byte) (string, error) {
...
+		str := ""
+		for x := 0; x < 56; x++ {
+			if x < len(s) {
+				str += string(s[x])
+			} else {
+				str += "="
+			}
+		}
```
Stebalien:
```go
if len(s) < 56 {
  s += strings.Repeat("=", len(s) - 56)
}

// or

for len(s) < 56 {
  s += "="
}
```

transcoders.go
```diff
@@ -238,13 +238,66 @@ func garlic64BtS(b []byte) (string, error) {
...
+	}
+	return garlicHostBytes, nil
+}
+
+func garlic32BtS(b []byte) (string, error) {
+	if len(b) > 35 {
+		if len(b) < 32 {
+			return "", fmt.Errorf("failed to validate garlic addr: %s not an i2p base32 address. len: %d\n", b, len(b))
+		}
+	}
```
Stebalien: Again, let's explain why.

transcoders.go
```diff
@@ -238,13 +238,66 @@ func garlic64BtS(b []byte) (string, error) {
...
+	unpad := func(s2 string) string {
+		return strings.Replace(s2, "=", "", -1)
```
Stebalien: `strings.TrimRight`?

transcoders.go
```diff
@@ -238,13 +238,66 @@ func garlic64BtS(b []byte) (string, error) {
...
+	}
+	return unpad(garlicBase32Encoding.EncodeToString(b)), nil
+}
+
+func garlic32Validate(b []byte) error {
+	if len(b) > 35 {
+		if len(b) < 32 {
+			return fmt.Errorf("failed to validate garlic addr: %s not an i2p base32 address. len: %d\n", b, len(b))
+		}
+	}
```
Stebalien: Comment.

transcoders.go
eyedeekay: It was wrong anyway, thanks! Fixed in the new changes.

transcoders.go
eyedeekay: Yeah not sure where my head was at here. Fixed.

transcoders.go
eyedeekay: Something seems to be off with my math. Just a moment.

eyedeekay: The reason why was that the padding required is could be shorter, longer, or 0 based, with 4 for regular .b32.i2p addresses, zero for 56 char encrypted leaseset .b32.i2p, or variable non-zero lengths for addresses with secrets. But then I forgot to do that.

transcoders.go
```diff
@@ -231,20 +232,75 @@ func garlic64StB(s string) ([]byte, error) {
...
+			// all other base32 addresses will always be exactly 52 characters
+			return nil, fmt.Errorf("failed to parse garlic addr: %s not a i2p base32 address. len: %d", s, len(s))
+		}
+	}
+	//compute the length to pad the address to, usually 56 or 64
+	padout := func(s string) string {
+		if len(s)%8 == 0 {
+			return s
+		}
+		x := int((len(s)/8)+1) * 8
```
Stebalien: You shouldn't need the `int` cast.

transcoders.go
```diff
@@ -231,20 +232,75 @@ func garlic64StB(s string) ([]byte, error) {
...
+		for len(s) < x {
```
Stebalien:
```go
for len(s) % 8 != 0 {
...
}
```

transcoders.go
```diff
@@ -231,20 +232,75 @@ func garlic64StB(s string) ([]byte, error) {
...
+	}
+	return strings.TrimRight(garlicBase32Encoding.EncodeToString(b), "="), nil
+}
+
+func garlic32Validate(b []byte) error {
+	// an i2p base64 for an Encrypted Leaseset v2 will be at least 35 bytes
+	// long
+	if len(b) < 35 {
+		// other than that, they will be exactly 32 bytes
+		if len(b) != 32 {
```
Stebalien: Let's combine these: `if len(b) < 35 && len(b) != 32 {` ...

Same with the other cases.

transcoders.go
```diff
@@ -231,20 +232,75 @@ func garlic64StB(s string) ([]byte, error) {
...
+	// an i2p base32 address with a length of greater than 55 characters is
+	// using an Encrypted Leaseset v2.
+	if len(s) < 55 {
+		if len(s) != 52 {
+			// all other base32 addresses will always be exactly 52 characters
+			return nil, fmt.Errorf("failed to parse garlic addr: %s not a i2p base32 address. len: %d", s, len(s))
+		}
+	}
+	//compute the length to pad the address to, usually 56 or 64
+	padout := func(s string) string {
```
Stebalien: (I'm still not sure why we need a function instead of just running this code)

transcoders.go
eyedeekay: It needs to be truncated to the whole digit?

transcoders.go
eyedeekay: Oh I see now.

transcoders.go
eyedeekay: Resolved. I hadn't thought of doing: 

``` Go
for len(s) % 8 != 0 {
    s += "="
}
```

to ensure the padding length.

PR #102, opened 2019-04-25
memoize string representation of multiaddrs

raulk:
Calculating the string representation of multiaddrs is costly; we can memoise it.

Here's the thing, though. multiaddrs are immutable and our Stringer functions are pure, so there no functional need to synchronise. I much have unguarded state and calculate the string twice, rather than synchronize on every access, but I don't think it's possible to pacify the Go race detector selectively.

In my benchmarks, `sync.Mutex` was 10x slower than `atomic.Value` which doesn't provide any atomicity actually (we can still calculate the string twice), but in terms of quietening the race detector, it's the cheapest option.

Cost of locking and unlocking a `sync.Mutex`:

```
BenchmarkMu-8   	100000000	        14.6 ns/op
```

Cost of storing a string once in an `atomic.Value`, and loading it N times:

```
BenchmarkAtomicValue-8   	2000000000	         1.32 ns/op
```

Baseline (calculating the string every time):

```
BenchmarkStringEverytime-8   	 1000000	      1760 ns/op
```

### DIFF
--- a/multiaddr.go
+++ b/multiaddr.go
@@ -6,11 +6,13 @@ import (
 	"fmt"
 	"log"
 	"strings"
+	"sync/atomic"
 )
 
 // multiaddr is the data structure representing a Multiaddr
 type multiaddr struct {
 	bytes []byte
+	str   atomic.Value
 }
 
 // NewMultiaddr parses and validates an input string, returning a *Multiaddr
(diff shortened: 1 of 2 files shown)


vyzo:
This might have deleterious effect in terms of peerstore memory usage. We do log pretty much all addrs somewhere, so these string representations can accumulate (and more than double the memory cost of the addr).
Not binding, but food for thought.

vyzo:
So it might be preferable to take the cost of recomputing the string repr as needed as opposed to caching it and increasing our memory usage by a factor of 2. 
It's space -vs- time tradeoff.

vyzo:
Has string repr calcuation come up anywhere in profiles as a problem?

raulk:
@vyzo came up here in discussion: https://github.com/libp2p/go-libp2p-swarm/pull/121/files#r278346429

I don't think we've kept an eye on heap profiles for this one. Perhaps the relay infrastructure can give us a data point?

Re: string retention, I agree. I just wanted to float the idea of memoising, not married to the idea; we need to quantify the impact before we take a decision.

vyzo:
The peerstore is the biggest accounted user of memory in relays (not a runaway leader, just the biggest user).

Stebalien:
Can we just avoid formatting multiaddrs? That is, we should really work towards fixing our logging framework so it doesn't format anything unless we have logging enabled.

(and @vyzo is correct about memory usage, we store a lot of multiaddrs)

raulk:
Memoizing is not the right solution. Thanks for the discussion!

PR #103, opened 2019-05-17
improve parse error

Stebalien:
This includes the bad multiaddr in the error and should improve UX (see https://github.com/ipfs/go-ipfs/issues/4190).

### DIFF
--- a/codec.go
+++ b/codec.go
@@ -7,7 +7,6 @@ import (
 )
 
 func stringToBytes(s string) ([]byte, error) {
-
 	// consume trailing slashes
 	s = strings.TrimRight(s, "/")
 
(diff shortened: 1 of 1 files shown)


codec.go
```diff
@@ -46,7 +45,7 @@ func stringToBytes(s string) ([]byte, error) {
 
 		a, err := p.Transcoder.StringToBytes(sp[0])
 		if err != nil {
-			return nil, fmt.Errorf("failed to parse %s: %s %s", p.Name, sp[0], err)
+			return nil, fmt.Errorf("failed to parse multiaddr %q: invalid argument %q for protocol %s: %s", s, sp[0], p.Name, err)
```
lgierth: They're called value, not argument :)

Stebalien: Fixed.

PR #105, opened 2019-05-20
forbid empty multiaddrs

### LINKED ISSUE #104: Reject empty multiaddrs
These are almost always useless and can cause bugs like https://github.com/ipfs/go-ipfs/issues/6354.

However, this would break the marshaling code as that expects to be able to construct these. It also removes the base-case for a recursive multiaddr definition...

cc @hsanjuan, @lgierth thoughts?

Stebalien:
fixes #104

### DIFF
--- a/multiaddr_test.go
+++ b/multiaddr_test.go
@@ -377,8 +377,8 @@ func TestEncapsulate(t *testing.T) {
 
 	m4, _ := NewMultiaddr("/ip4/127.0.0.1")
 	d := c.Decapsulate(m4)
-	if s := d.String(); s != "" {
-		t.Error("decapsulate /ip4 failed.", "/", s)
+	if d != nil {
+		t.Error("decapsulate /ip4 failed: ", d)
 	}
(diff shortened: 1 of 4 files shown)


multiaddr_test.go
```diff
@@ -582,11 +582,11 @@ func TestBinaryMarshaler(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	addr2 := newMultiaddr(t, "")
+	var addr2 multiaddr
```
Stebalien: This is the issue I was talking about. This could be done by manually calling `NewMultiaddr` before but now one needs to construct a _valid_ multiaddr.

hsanjuan: but would work by creating a non-empty multiaddress (which just will be replaced) ?

Stebalien: Yes, I guess. Although... multiaddrs are really supposed to be immutable. I guess this is just an unavoidable edge-case.

PR #106, opened 2019-05-31
Implement memory multiaddrs

bigs:
Implements memory multiaddrs with uint64 parameters.


lidel:
Continued in https://github.com/multiformats/go-multiaddr/pull/256

Issue #108, opened 2019-07-18
Move all protocol definitions into this package


Stebalien:
Due to https://github.com/multiformats/multiaddr/issues/70, we can't handle mulitaddrs we don't understand. For now, we should just understand them all.

Stebalien:
Specifically, move:
* p2p-circuit from go-libp2p-circuit
* ws/wss from go-ws-transport
* dns{4,6,addr} from go-multiaddr-dns (not the entire package, _just_ the protocol definitions).

PR #111, opened 2019-09-18
fix https://github.com/multiformats/go-multiaddr/issues/108

godcong:
1. move p2p-circuit protocol definitions(https://github.com/libp2p/go-libp2p-circuit/blob/master/transport.go#L13-L24)
2. move ws protocol definitions(https://github.com/libp2p/go-ws-transport/blob/ec8d1818a2e6edad2496410cbc554df6fe930ea2/websocket.go#L23)
3. move dns protocol definitions(https://github.com/multiformats/go-multiaddr-dns/blob/3974bf3f84c52825588fdcc0fd0e0aa7953ab5ff/dns.go#L11)

to fix https://github.com/multiformats/go-multiaddr/issues/108

### DIFF
--- a/protocols.go
+++ b/protocols.go
@@ -5,6 +5,9 @@ package multiaddr
 const (
 	P_IP4               = 0x0004
 	P_TCP               = 0x0006
+	P_DNS4              = 0x0036
+	P_DNS6              = 0x0037
+	P_DNSADDR           = 0x0038
 	P_UDP               = 0x0111
 	P_DCCP              = 0x0021
 	P_IP6               = 0x0029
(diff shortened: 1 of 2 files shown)


godcong:
if use this change, need remove the init of the previous packages.
if not do will:
panic: error registering dns4 protocol: protocol by the name "dns4" already exists

goroutine 1 [running]:
github.com/multiformats/go-multiaddr-dns.init.0()
	github.com/multiformats/go-multiaddr-dns@v0.0.2/dns.go:42 +0x221

Stebalien:
Thanks! I'll go ahead and update those packags.

PR #112, opened 2019-09-18
fix: minimal varint decoding

Stebalien:
* Enforce minimal varint encoding when decoding multiaddrs. See https://github.com/multiformats/unsigned-varint/pull/19
* Ensure codecs fit in 32bit integers for compatibility. We can increase the size, but we'd need to change the `Protocol.Codec` type from `int` to `uint64`.


varint.go
```diff
@@ -1,44 +1,27 @@
...
-}
+	"github.com/multiformats/go-varint"
+)
 
 // CodeToVarint converts an integer to a varint-encoded []byte
 func CodeToVarint(num int) []byte {
-	buf := make([]byte, VarintSize(num))
-	n := binary.PutUvarint(buf, uint64(num))
-	return buf[:n]
+	if num < 0 || num > math.MaxInt32 {
```
bigs: maybe we add a helper function that accepts a `uint32` and start to prefer using that directly, but have this function execute these guards and then attempt to cast and call that helper function

Stebalien: Well, we _should_ have one that accepts uint64. But we don't need to tackle that yet.

Stebalien:
@bigs: Note: I believe we need to enforce this in other places (peer IDs, multicodecs, etc.)

PR #116, opened 2019-12-09
Validate bytes when parsing string

aratz-lasa:
If no validation is done, it can create invalid Multiaddr.
Probably there is an obvious reason for not doing it. So, if it is the case, I am sorry.
As an example, you can create an ip6zone Multiaddr from the following string: `ip6zone/`. However, if you validate it, it raises an error, because it contains `/`

### DIFF
--- a/transcoders.go
+++ b/transcoders.go
@@ -14,8 +14,11 @@ import (
 )
 
 type Transcoder interface {
+	// Validates and encodes to bytes a multiaddr that's in the string representation.
 	StringToBytes(string) ([]byte, error)
+	// Validates and decodes to a string a multiaddr that's in the bytes representation.
 	BytesToString([]byte) (string, error)
+	// Validates bytes when parsing a multiaddr that's already in the bytes representation.
 	ValidateBytes([]byte) error
 }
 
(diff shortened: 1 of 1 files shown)


Stebalien:
`ip6zoneStB` should validate the string internally. The `ValidateBytes` function is meant for validating bytes when parsing a multiaddr that's already in the bytes representation.

Could you change the patch to call the `ip6zoneVal` function from `ip6zoneStB`?

(we should also document this)

Stebalien:
`ip6zoneStB` should validate the string internally. The `ValidateBytes` function is meant for validating bytes when parsing a multiaddr that's already in the bytes representation.

Could you change the patch to call the `ip6zoneVal` function from `ip6zoneStB`?

(we should also document this)

aratz-lasa:
Okay, understood. Maybe it is better to call directly `strings.Contains(s, "/")`? 
Where should we document it?

Stebalien:
> Maybe it is better to call directly `strings.Contains(s, "/")`? 

SGTM.

> Where should we document it?

Let's document the functions on the `Transcoder` interface.

aratz-lasa:
Okay, done. I hope it is good enough, tell me if you want to change something.

Stebalien:
Thanks!

PR #117, opened 2019-12-13
don't return leading / for unix values on Windows

djdv:
This changes Unix components to return a valid/native system path on Windows from their `.Value` method.

Consider a multiaddr constructed from `/unix/c:\Users\me\unix.socket`.

Currently, the value returned is `/c:\Users\me\unix.socket`. Which isn't very useful.
This change returns `c:\Users\me\unix.socket` instead, so that the path value may be passed directly from the multiaddr component to other system functions (within `os`, `net`, etc.).

I figured it made more sense for `component.Value()` to return a system native value from the source rather than expect users to parse the output and have special platform handling in their codebase. 
```go
multiaddr.ForEach(ma, func(comp multiaddr.Component) bool {
	if comp.Protocol().Code == multiaddr.P_UNIX {
		// assume multiaddr package will return a native value
		os.Remove(comp.Value())
...
```
vs
```go
multiaddr.ForEach(ma, func(comp multiaddr.Component) bool {
	if comp.Protocol().Code == multiaddr.P_UNIX {
		val := comp.Value()
		if runtime.GOOS == "windows" {
			val = strings.TrimLeft(val, "/"))
		}
		// more special cases would be needed here later
		os.Remove(val)
...
```

### DIFF
--- a/transcoders.go
+++ b/transcoders.go
@@ -7,6 +7,7 @@ import (
 	"encoding/binary"
 	"fmt"
 	"net"
+	"runtime"
 	"strconv"
 	"strings"
 
@@ -317,10 +318,21 @@ func p2pBtS(b []byte) (string, error) {
 var TranscoderUnix = NewTranscoderFromFunctions(unixStB, unixBtS, nil)
 
 func unixStB(s string) ([]byte, error) {
+	if runtime.GOOS == "windows" {
+		return []byte(strings.TrimLeft(s, "/")), nil
+	}
 	return []byte(s), nil
 }
 
 func unixBtS(b []byte) (string, error) {
+	if runtime.GOOS == "windows" {
+		if len(b) < 1 {
+			return "", nil
+		}
+		if b[0] == '/' {
+			return string(b[1:]), nil
+		}
+	}
 	return string(b), nil
 }
 


djdv:
Note, this breaks the tests:
`multiaddr_test.go:402: expected "/a/b/c/d" for 400 in /ip4/0.0.0.0/unix/a/b/c/d, but got "a/b/c/d" instead`

But these trimmed value's are still equivalent in this context.
```go
ca, _ := multiaddr.NewComponent("unix", "a/b/c")
cb, _ := multiaddr.NewComponent("unix", "/a/b/c")
ca.Equal(cb) == true
```

Whether this breaks expectations or is considered acceptable for the pkg though, I'm not sure.
If this seems like a sensible approach I can look into changing the test, but if these values are expected to be exact, maybe it would make more sense to add a separate method (`NativeValue()`?) or just expect users to handle platform differences (not ideal imo).

Stebalien:
This is tricky, I'm not sure if we can really solve this.

* Paths on _unix_ should start with /.
* `/unix` multiaddrs should use forward slashes on all platforms, both in text and in binary. For completeness, we'd have to convert paths to forward slashes when we encode them.
* We should convert back to backslashes on Windows. However, from the perspective of _this_ package, the "string" value is the value as seen in a string multiaddr. If we change `unixBtS` to convert slashes, we'd break our string formatting of multiaddrs.

Given that, I don't think we can really fix this. We should tell users to use go-multiaddr-net (you solved the same issue there) or to convert paths as necessary.

PR #122, opened 2020-02-29
feat: add wss protocol

Stebalien:
Now that we define all protocols in this package, we might as well add wss.

fixes #72


Stebalien:
@raulk I'm requesting a postmergus review as this shouldn't be controversial.

PR #125, opened 2020-05-20
feat: move filters into a sub-package

Stebalien:
These really don't belong in the root package.

### DIFF
--- a/filter/filter.go
+++ b/filter/filter.go
@@ -1,8 +1,10 @@
-package multiaddr
+package filter
 
 import (
 	"net"
 	"sync"
+
+	ma "github.com/multiformats/go-multiaddr"
 )
 
 // Action is an enum modelling all possible filter actions.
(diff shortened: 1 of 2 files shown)


Stebalien:
I agree that single-type packages are suboptimal, but it _doesn't_ belong in the root package. A multiaddr _set_ might belong in the root package, but this filter type is highly specialized (special IPNet functions, filter actions, etc.). Remember, this package has implementations in _many_ languages and needs to present a cohesive minimal interface.

How about a compromise:

1. Move `go-multiaddr-net` into `go-multiaddr/net`.
2. Move this filter package into `go-multiaddr/net`.

raulk:
@Stebalien Deal taken! That simplifies our codebase even further. Is this something you plan on tackling?

Stebalien:
Yeah. It's been annoying me for a while... I wonder if I can fold go-multiaddr-dns into the deal.

raulk:
@Stebalien Nowwwww we're talking :-)

Stebalien:
Replaced with #127.

PR #130, opened 2020-08-11
Support /p2p addresses encoded as CIDs

aschmahmann:
This adds go-cid as a dependency of this library. If adding go-cid is contentious lmk and I can try and copy-paste the parts we need.

Not sure why codecov says the new code isn't tested. Pretty sure I've added enough new test cases. Extra eyes appreciated though.


transcoders.go
```diff
@@ -294,11 +295,15 @@ var TranscoderP2P = NewTranscoderFromFunctions(p2pStB, p2pBtS, p2pVal)
 
 func p2pStB(s string) ([]byte, error) {
 	// the address is a varint prefixed multihash string representation
-	m, err := mh.FromB58String(s)
+	c, err := cid.Decode(s)
```
Stebalien, line 298: This won't work for ed25519 keys.

aschmahmann: Lol, well I should've gotten that. 🤦

Stebalien: Note: I think you can also check "does it start with 1 or Q". If so, decode from base58.

transcoders.go
```diff
@@ -293,12 +294,22 @@ func garlic32Validate(b []byte) error {
...
+		return m, nil
+	}
+
+	// check if the address is a CID
+	c, err := cid.Decode(s)
 	if err != nil {
 		return nil, fmt.Errorf("failed to parse p2p addr: %s %s", s, err)
 	}
-	return m, nil
+	if c.Version() > 0 && c.Type() != cid.Libp2pKey {
```
Stebalien, line 308: `c.Version()` should be 1, not o.

transcoders.go
aschmahmann: Sure, I can just copy the code from https://github.com/libp2p/go-libp2p-core/blob/b77fd280f2bfcce22f10a000e8e1d9ec53c47049/peer/peer.go#L172, but that's technically a spec violation.

transcoders.go
aschmahmann: I have `c.Version() > 0` here. I guess we don't know what CIDv2+ will look like anyway so I can just assert `c.Version() == 1` though.

aschmahmann: I decided to just ignore the Version field entirely as we do in go-libp2p-core. As far as I can tell there is literally no way (except perhaps with reflection) to create a CID with version number not equal to 0 or 1.

Lmk if there are any other checks you think are missing here.

transcoders.go
```diff
@@ -292,13 +293,34 @@ func garlic32Validate(b []byte) error {
...
+
+	ty := c.Type()
+	if ty == cid.Libp2pKey {
+		return c.Hash(), nil
+	}
+
+	codecStr := cid.CodecToStr[ty]
+	if codecStr == "" {
+		codecStr = fmt.Sprintf("[unknown multicodec %d]", ty)
+	}
```
aschmahmann, line 322: I copied this from go-libp2p-core, but is this overkill?

multiaddr_test.go
```diff
@@ -558,6 +575,31 @@ func TestInvalidP2PAddr(t *testing.T) {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// using MD5 since it's not a valid data codec
+	unknownCodecCID := cid.NewCidV1(mh.MD5, hashedData).String()
+
+	badStringAddrs := []string{
+		"/p2p/k2k4r8oqamigqdo6o7hsbfwd45y70oyynp98usk7zmyfrzpqxh1pohl-", // invalid multibase encoding
+		"/p2p/?unknownmultibase", // invalid multibase encoding
```
aschmahmann, line 589: Is this a safe/reasonable test? I guess we can just remove it if we ever allocate the `?` multibase and this test fails.

transcoders.go
Stebalien: Ah, nvm, now I get it.

> I decided to just ignore the Version field entirely as we do in go-libp2p-core. As far as I can tell there is literally no way (except perhaps with reflection) to create a CID with version number not equal to 0 or 1.

Checking the type should be sufficient.

transcoders.go
Stebalien: Yeah, this seems like overkill. We can just fail.

multiaddr_test.go
Stebalien: Seems reasonable.

multiaddr_test.go
```diff
@@ -558,6 +575,31 @@ func TestInvalidP2PAddr(t *testing.T) {
...
 }
 
+func TestInvalidP2PAddrString(t *testing.T) {
+	hashedData, err := mh.Sum([]byte("test"), mh.SHA2_256, -1)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// using MD5 since it's not a valid data codec
+	unknownCodecCID := cid.NewCidV1(mh.MD5, hashedData).String()
```
aschmahmann, line 585: Maybe an overkill test, but testing for an invalid codec.

transcoders.go
aschmahmann: removed, look good?

aschmahmann:
btw @Stebalien any idea how I managed to decrease code coverage in this one area https://codecov.io/gh/multiformats/go-multiaddr/compare/2da13ecd1ade1ef08fc453f1b2527bd4edb5ec98...9eb5a88b0945beda7cee97da5086e82acd1a21ef/changes?

Stebalien:
I have no idea. Codecov is weird.

PR #132, opened 2020-08-20
dep: add "codependencies" for handling version conflicts

Stebalien:
This patch adds a special, internal, "codependencies" package that allows us to specify minimum required versions for modules we don't directly depend on.

Problem: Updating go-multiaddr without updating go-multiaddr-net and/or go-maddr-filter can cause conflicting type issues.

Proposal: Add "dependencies" on these _modules_ (circular module dependencies are perfectly legal), but avoid actually importing any packages.

Downside:

* Anyone using this module will also have to download these deprecated modules.

Upside:

* After upgrading this module, all "codependencies" will automatically be upgraded to compatible versions.
* If unused, codependencies will _not_ end up bloating the binary size. This is because "usage" is computed at the import/package layer, while module version requirements are computed at the module layer.

### DIFF
--- a/internal/codependencies/README.md
+++ b/internal/codependencies/README.md
@@ -2,6 +2,8 @@ This package allows us to depend on other modules, to _force_ these other
 modules to upgrade to some minimum version. This allows us to express "co
 dependency" requirements in cases where this module doesn't strictly speaking
 _depend_ on another module, but conflicts with some version of that module.
+We are using this here to depend on deprecated modules that have been
+merged into this package. 
 
 In practice, this means:
 
--- a/internal/codependencies/README.md
+++ b/internal/codependencies/README.md
@@ -1,0 +1,12 @@
+This package allows us to depend on other modules, to _force_ these other
+modules to upgrade to some minimum version. This allows us to express "co
+dependency" requirements in cases where this module doesn't strictly speaking
+_depend_ on another module, but conflicts with some version of that module.
+
+In practice, this means:
+
+1. Packages imported here _will not_ end up in the final binary as nothing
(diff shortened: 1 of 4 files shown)


Stebalien:
Notes:

* The "co-dependencies" are actually just shims, in this case, so it doesn't add too much weight.
* I consider this to be a temporary upgrade measure. After some period of time, we should probably remove these.

internal/codependencies/README.md
```diff
@@ -0,0 +1,12 @@
...
+dependency" requirements in cases where this module doesn't strictly speaking
+_depend_ on another module, but conflicts with some version of that module.
+
+In practice, this means:
+
+1. Packages imported here _will not_ end up in the final binary as nothing
+   imports this package.
+2. Modules containing these packages will, unfortunately, be downloaded as
+   "dependencies" of this package.
+3. Anyone using this module will be forced to upgrade all co-dependencies.
```
raulk, line 12: It's not immediately clear from this README.md that these imports are actually deprecated and absorbed into go-multiaddr. Can we make this clear in the README.md?

Anecdote: I had forgotten we had (finally) consolidated these libs, and was about to review this with "if we're needing to do this, it's a smell that those libraries should actually live here."

raulk (commented):
Hmmm... IMO the need to do this is a smell that those libraries should live in this repo, under different

internal/codependencies/README.md
aschmahmann: @raulk filter is not yet moved in here, only net is

Stebalien:
Filter has moved here as well.

PR #135, opened 2021-01-21
Fix Local Address on TCP connections

aarshkshah1992:
Closes https://github.com/libp2p/go-libp2p/issues/1024.

### DIFF
--- a/net/net.go
+++ b/net/net.go
@@ -255,7 +255,15 @@ func (l *maListener) Accept() (Conn, error) {
 		}
 	}
 
-	return wrap(nconn, l.laddr, raddr), nil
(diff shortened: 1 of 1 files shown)


willscott:
it seems somewhat awkward that the `laddr` on the listener is different from the local address that's actually used on connections. Is the issue that there can be multiple potential local addresses for a single listener?

Stebalien (approved):

Stebalien:
> it seems somewhat awkward that the laddr on the listener is different from the local address that's actually used on connections. Is the issue that there can be multiple potential local addresses for a single listener?

Yes. E.g., we could be listening on the "zero" address.

PR #152, opened 2021-05-24
fix(net): export new net.Addr conversion registration functions

### LINKED ISSUE #151: Registering new multicodecs
You can see here a wonderful, not-exported singleton registry variable that is in this delightful in-between where it's deprecated but functionality has not been maintained:

https://github.com/multiformats/go-multiaddr/blob/master/net/registry.go#L17

I get these deprecated markups in my IDE because the only way to add new codecs to this is marked deprecated, and it directs you to two functions *that depend on you first being able to **address** the aforementioned singleton registry variable*.

I only learned about this by looking at an old, not-updated uTP transport implementation. If it were my repository that singleton would be exported already, and the deprecation warnings would have some actual meaning. Or at least, to maintain some semblance of extensibility.

It's not that it's not extensible, it's just marked deprecated where you would extend it.

I figure first start the conversation, I am probably missing something as dealing with multiformats and libp2p involves a lotta lotta individual repositories which can be quite tedious to trace back and forth.

I am here mainly because I want to add a new protocol, namely https://github.com/xtaci/kcp and to do this it really needs to have kcp in there (yes I have already submitted a PR to add it to the CSV). This is day 3 of me tackling this project, I'm starting to make sense of it all but it's tricky.

I don't mind what way it gets resolved, for now I am just gonna smile at the strikethroughs in my syntax highlighting.

Stebalien:
fixes #151

### DIFF
--- a/net/registry.go
+++ b/net/registry.go
@@ -17,13 +17,13 @@ type ToNetAddrFunc func(ma ma.Multiaddr) (net.Addr, error)
 var defaultCodecs = NewCodecMap()
 
 func init() {
-	defaultCodecs.RegisterFromNetAddr(parseTCPNetAddr, "tcp", "tcp4", "tcp6")
-	defaultCodecs.RegisterFromNetAddr(parseUDPNetAddr, "udp", "udp4", "udp6")
-	defaultCodecs.RegisterFromNetAddr(parseIPNetAddr, "ip", "ip4", "ip6")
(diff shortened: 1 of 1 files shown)


marten-seemann (approved):
LGTM, but CI is unhappy. For some reason, the Codecov checksum verification failed.

Stebalien:
:man_shrugging: We'll see if that becomes a "thing".

PR #153, opened 2021-05-26
Add TLS protocol

carpawell:
Added TLS protocol. There are plenty of applications of TLS protocol not in the context of using it with `http` or `ws`(e.g. in `gPRC`), so it seems logical to add it. Code for it was taken from [multicodec](https://github.com/multiformats/multicodec) repo.

There is https://github.com/multiformats/multiaddr/pull/109 PR that suggests adding TLS protocol, mark `https` and `wss` protocols as deprecated and start using `/tls/http` and `/tls/ws` instead.

### DIFF
--- a/multiaddr_test.go
+++ b/multiaddr_test.go
@@ -140,7 +140,7 @@ func TestConstructSucceeds(t *testing.T) {
 		"/udp/1234/udt",
 		"/udp/1234/utp",
 		"/tcp/1234/http",
-		"/tcp/1234/http/tls",
+		"/tcp/1234/tls/http",
 		"/tcp/1234/https",
 		"/ipfs/QmcgpsyWgH8Y8ajJz1Cu72KnS5uo2Aa2LpzU7kinSupNKC/tcp/1234",
 		"/ipfs/k2k4r8oqamigqdo6o7hsbfwd45y70oyynp98usk7zmyfrzpqxh1pohl7/tcp/1234",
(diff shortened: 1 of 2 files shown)


multiaddr_test.go
```diff
@@ -140,6 +140,7 @@ func TestConstructSucceeds(t *testing.T) {
 		"/udp/1234/udt",
 		"/udp/1234/utp",
 		"/tcp/1234/http",
+		"/tcp/1234/http/tls",
```
Stebalien, line 143:
```suggestion
		"/tcp/1234/tls/http",
```

multiaddr_test.go
```diff
@@ -168,6 +169,7 @@ func TestConstructSucceeds(t *testing.T) {
 		"/ip4/127.0.0.1/tcp/9090/http/p2p-webrtc-direct",
 		"/ip4/127.0.0.1/tcp/127/ws",
 		"/ip4/127.0.0.1/tcp/127/ws",
+		"/ip4/127.0.0.1/tcp/127/ws/tls",
```
Stebalien, line 172:
```suggestion
		"/ip4/127.0.0.1/tcp/127/tls/ws",
```

Stebalien (approved):
As long as we're not simply _removing_ /https, this seems reasonable. Let me quickly get a signoff from other libp2p devs but then I think we can probably just merge it.

mxinden (approved):
This looks good to me from the multiformats perspective. I can not comment on the Go side.

I think we should wait for a sign-off by @marten-seemann as well to ensure this is consistent long term with QUIC and HTTP3.

marten-seemann (approved):

PR #155, opened 2021-06-23
guard against nil {Local,Remote}Addr() return values

marten-seemann:
Fixes https://github.com/ipfs/go-ipfs/issues/8211. As discussed in https://github.com/ipfs/go-ipfs/issues/8211#issuecomment-867036960, I'm not sure why `LocalAddr()` of a `net.Conn` would ever return `nil`, but as evidenced by the stack trace, it can happen (maybe just on certain architectures).

Playing it safe, we can check that the value is not `nil` before using it.


Stebalien (approved):
This is strange... but yeah, we might as well do it.

PR #160, opened 2021-07-24
add functions to tell if an address contains a security protocol

marten-seemann:


multiaddr.go
```diff
@@ -184,3 +184,14 @@ func (m *multiaddr) ValueForProtocol(code int) (value string, err error) {
 	})
 	return
 }
+
+func ContainsSecurityProtocol(addr Multiaddr) bool {
+	var contains bool
+	ForEach(addr, func(c Component) bool {
+		if code := c.Protocol().Code; code == P_TLS || code == P_NOISE {
```
Stebalien, line 191: - I would definitely put this in an extendable map.
- This is also somewhat ambiguous. For example, what about `wss`? What about `quic`? This kind of check will likely fit better in the swarm itself.

marten-seemann: > * I would definitely put this in an extendable map.

Makes sense.

> * This is also somewhat ambiguous. For example, what about wss? What about quic? This kind of check will likely fit better in the swarm itself.

True, this should probably be better documented.
I'm not sure if putting it in the swarm is the right solution though. We'll need a function like this (maybe a `SecurityProtocol(Multiaddr) string` is even better?) in quite a lot of packages, especially as we'll have a lengthy transition period.

PR #162, opened 2021-07-25
remove forced dependency on deprecated go-maddr-filter

marten-seemann:
I'm happy to get rid of this hack.


Stebalien (approved):
It's been a year.

PR #168, opened 2021-12-14
move ResolveUnspecifiedAddress(es) and FilterAddrs here from libp2p/go-addr-util

marten-seemann:
These are the only functions in use from libp2p/go-addr-util. By moving them here, we can deprecate that package.

### DIFF
--- a/net/ip.go
+++ b/net/ip.go
@@ -95,8 +95,8 @@ func IsIPUnspecified(m ma.Multiaddr) bool {
 	return net.IP(c.RawValue()).IsUnspecified()
 }
 
-// AddrOverNonLocalIP returns whether the addr uses a non-local ip link
-func AddrOverNonLocalIP(a ma.Multiaddr) bool {
+// IsIpv6LinkLocal returns whether the addr uses a non-local ip link
+func IsIpv6LinkLocal(a ma.Multiaddr) bool {
 	split := ma.Split(a)
 	if len(split) < 1 {
 		return false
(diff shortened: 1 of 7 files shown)


vyzo (approved):

net/ip.go
```diff
@@ -95,6 +95,18 @@ func IsIPUnspecified(m ma.Multiaddr) bool {
 	return net.IP(c.RawValue()).IsUnspecified()
 }
 
+// AddrOverNonLocalIP returns whether the addr uses a non-local ip link
+func AddrOverNonLocalIP(a ma.Multiaddr) bool {
```
Stebalien, line 99: This is a strange function. What about IPv4 localhost?

Stebalien (approved):

net/ip.go
marten-seemann: This is where we use it: https://github.com/libp2p/go-libp2p-swarm/blob/69f15a70a0caa2ad4f94bfc603f63181b13a40fb/swarm_dial.go#L636-L661

Is there any reason we don't want to dial IPv6 link-local addresses, but are ok with dialing IPv4 localhost?
At least the code here is consistent with the comment in the swarm.

Stebalien: > Is there any reason we don't want to dial IPv6 link-local addresses, but are ok with dialing IPv4 localhost?

Because they're almost always useless while localhost can actually be useful (and is usually only one address). NOTE: I don't think we _advertise_ link-local addresses anymore.

Really, this _isn't_ the right place to do things like this, but it's the only place till we get a real address "scope" system.

>  At least the code here is consistent with the comment in the swarm.

I'd change it to `IsIpv6LinkLocal`, or something like that.

marten-seemann: Makes sense, I renamed the function. All (the few) users of that functions will need to adjust the path manually anyway, so this doesn't create too much extra hassle.

PR #170, opened 2021-12-21
remove wrong (and redundant) IsIpv6LinkLocal

### LINKED ISSUE #169: wrong IsIpv6LinkLocal
This change https://github.com/multiformats/go-multiaddr/pull/168#discussion_r772750843 was wrong. It's doing the exact opposite of what the name suggests.

marten-seemann:
Fixes #169.

We should just use `IsIP6LinkLocal` instead.

### DIFF
--- a/net/resolve.go
+++ b/net/resolve.go
@@ -74,7 +74,7 @@ func interfaceAddresses() ([]ma.Multiaddr, error) {
 
 	var out []ma.Multiaddr
 	for _, a := range maddrs {
-		if !IsIpv6LinkLocal(a) {
+		if IsIP6LinkLocal(a) {
 			continue
 		}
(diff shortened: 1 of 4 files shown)


net/resolve.go
```diff
@@ -74,7 +74,7 @@ func interfaceAddresses() ([]ma.Multiaddr, error) {
 
 	var out []ma.Multiaddr
 	for _, a := range maddrs {
-		if !IsIpv6LinkLocal(a) {
+		if IsIP6LinkLocal(a) {
```
marten-seemann, line 77: Note that `IsIpv6Local` was doing the opposite of what the name suggested.

Stebalien (approved):

PR #173, opened 2022-01-10
release v0.5.1

marten-seemann:


marten-seemann:
Any idea why `gorelease` is suggestion v0.6.0 here? This should be a backwards-compatible change.

mvdan:
Because, as per semver, a bugfix release is not meant to add any API. This also applies to v0.

marten-seemann:
Then we should give this a bit more time, to avoid too much release churn.

PR #176, opened 2022-06-13
add WebTransport multiaddr components

marten-seemann:
Based on https://github.com/libp2p/specs/pull/404.


transcoders.go
```diff
@@ -373,3 +374,14 @@ func dnsStB(s string) ([]byte, error) {
...
+
+var TranscoderCertHash = NewTranscoderFromFunctions(certHashStB, certHashBtS, nil)
+
+func certHashStB(s string) ([]byte, error) {
+	_, data, err := multibase.Decode(s)
+	return data, err
+}
+
+func certHashBtS(b []byte) (string, error) {
+	return multibase.Encode(multibase.Base58BTC, b)
```
marten-seemann, line 386: This choice is totally arbitrary. We could use any other multibase that's widely supported.

Is there any best practice for choosing one over the other?

aschmahmann: cc @lidel, but the general guidelines I know of are:
- Do you want compact -> choose base64URL (`u`), although base58btc is probably fine too
- Do you want lowercase -> choose base32 (`b`)

transcoders.go
```diff
@@ -373,3 +374,14 @@ func dnsStB(s string) ([]byte, error) {
 func dnsBtS(b []byte) (string, error) {
 	return string(b), nil
 }
+
+var TranscoderCertHash = NewTranscoderFromFunctions(certHashStB, certHashBtS, nil)
+
+func certHashStB(s string) ([]byte, error) {
+	_, data, err := multibase.Decode(s)
```
aschmahmann, line 381: Are there any other requirements here? e.g. is it a multihash, is it just 32 bytes, etc.

multiaddr_test.go
```diff
@@ -157,6 +158,8 @@ func TestConstructSucceeds(t *testing.T) {
 		"/ip4/127.0.0.1/tcp/1234",
 		"/ip4/127.0.0.1/tcp/1234/",
 		"/ip4/127.0.0.1/udp/1234/quic",
+		"/ip4/127.0.0.1/udp/1234/quic/webtransport",
+		"/ip4/127.0.0.1/udp/1234/quic/webtransport/certhash/zt1Zv2yaZ",
```
aschmahmann, line 162: IIUC there can be multiple certhashes, if so we should probably test for multiple certhashes as well.

I'm not sure if the place is here or elsewhere, but wherever we're defining the semantics of `webtransport` and `certhash` we should add some tests around escaping (e.g. making sure you can do `.../webtransport/certhash/hash1/certhash/hash2/p2p/peerID` but not `.../webtransport/certhash/hash1/p2p/peerID/certhash/hash2`

marten-seemann: > IIUC there can be multiple certhashes, if so we should probably test for multiple certhashes as well.

Good point. Will add.

> I'm not sure if the place is here or elsewhere, but wherever we're defining the semantics of webtransport and certhash we should add some tests around escaping

It probably needs to live in the WebTransport implementation. Having one certhash, then something else, then another certhash can be a valid multiaddr (e.g. a circuit).

transcoders.go
marten-seemann: I added the validation on encoding.

transcoders.go
marten-seemann: Switched to `Base64url`.

mxinden (commented):
:rocket:

protocols.go
```diff
@@ -15,6 +15,8 @@ const (
 	P_IP6ZONE           = 0x002A
 	P_IPCIDR            = 0x002B
 	P_QUIC              = 0x01CC
+	P_WEBTRANSPORT      = 0x01D1
+	P_CERTHASH          = 0x01D2
```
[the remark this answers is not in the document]
marten-seemann: https://github.com/multiformats/multiaddr/pull/130

PR #179, opened 2022-07-01
Add webrtc

ckousik:
Adds an `x-webrtc` protocol which holds the 32 byte hex encoded DTLS fingerprint received by the server. This is based on https://github.com/melekes/rust-multiaddr/commit/c88f5f4cf1ad223fe4753f081a88d3ca31f44d63 and is for supporting the Go implementation of the spec described in https://github.com/libp2p/specs/pull/412 .

### DIFF
--- a/protocols.go
+++ b/protocols.go
@@ -36,7 +36,7 @@ const (
 	P_WS                = 0x01DD
 	P_WSS               = 0x01DE // deprecated alias for /tls/ws
 	P_PLAINTEXTV2       = 0x706c61
-	P_WEBRTC            = 0x115
+	P_WEBRTC            = 0x118
 )
 
 var (
--- a/protocols.go
+++ b/protocols.go
@@ -36,7 +36,7 @@ const (
 	P_WS                = 0x01DD
 	P_WSS               = 0x01DE // deprecated alias for /tls/ws
 	P_PLAINTEXTV2       = 0x706c61
-	P_XWEBRTC           = 0x115
+	P_WEBRTC           = 0x115
 )
 
 var (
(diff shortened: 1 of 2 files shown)


marten-seemann:
Wasn't the plan to use `/webrtc/certhash/xxx`?

mxinden:
> Wasn't the plan to use `/webrtc/certhash/xxx`?

:+1: Followed up on https://github.com/libp2p/specs/pull/412/files#r917573385.

protocols.go
```diff
@@ -247,6 +248,11 @@ var (
 		Code:  P_WSS,
 		VCode: CodeToVarint(P_WSS),
 	}
+	protoXWebRTC = Protocol {
```
marten-seemann, line 251: I might be missing something, what's the `X` doing here?
```suggestion
	protoWebRTC = Protocol {
```

marten-seemann: You also need to add this in `init` (right below).

ckousik: I missed this when removing the transcoders which are now unnecessary. Thank you!

marten-seemann (approved):

marten-seemann:
CI is unhappy about this PR. I think you'll need to run `gofmt`.

ckousik:
@marten-seemann Seems like `0x115` (277) is already used by `p2p-stardust` https://github.com/multiformats/multiaddr/blob/master/protocols.csv#L33 . Should I change the code for webrtc to 280 or some other nearby number?

marten-seemann:
Yes, makes sense. Just pick a number.

marten-seemann (approved):

PR #181, opened 2022-09-07
Fix comment on Decapsulate

MarcoPolo:
This comment was wrong.

### DIFF
--- a/interface.go
+++ b/interface.go
@@ -50,6 +50,8 @@ type Multiaddr interface {
 	// Decapsulate removes a Multiaddr wrapping. For example:
 	//
 	//      /ip4/1.2.3.4/tcp/80 decapsulate /tcp/80 = /ip4/1.2.3.4
+	//      /ip4/1.2.3.4/tcp/80 decapsulate /udp/80 = /ip4/1.2.3.4/tcp/80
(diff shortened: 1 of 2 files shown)


interface.go
```diff
@@ -47,9 +47,9 @@ type Multiaddr interface {
 	//
 	Encapsulate(Multiaddr) Multiaddr
 
-	// Decapsultate removes a Multiaddr wrapping. For example:
+	// Decapsulate removes a Multiaddr wrapping. For example:
 	//
-	//      /ip4/1.2.3.4/tcp/80 decapsulate /ip4/1.2.3.4 = /tcp/80
+	//      /ip4/1.2.3.4/tcp/80 decapsulate /tcp/80 = /ip4/1.2.3.4
```
marten-seemann, line 52: Should this also document what happens if you try to decapsulate something that's not there at the right hand side of the multiaddr (it returns `nil`).

marten-seemann (approved):

interface.go
MarcoPolo: It should! But the behavior is `/ip4/1.2.3.4/tcp/80 decapsulate /udp/80 = /ip4/1.2.3.4/tcp/80`

PR #193, opened 2023-03-01
Add percentencode impl

MarcoPolo:
In draft until I add the corresponding spec change


MarcoPolo:
I don't think we need this anytime soon. In libp2p+HTTP we have `/http-path`

PR #195, opened 2023-03-30
Update webrtc protocols after rename

MarcoPolo:
Context: https://github.com/multiformats/multiaddr/pull/152


marten-seemann (approved):
I was assuming LBL would do this...

Issue #198, opened 2023-04-18
implement a new multiaddress API


marten-seemann:
## Problems with the current API

Major problems:

- When receiving a multiaddress (e.g. from an address record in the DHT), we immediately try to parse it. If parsing fails, that multiaddress is discarded. Parsing fails if the multiaddress contains any unknown components: if a node hasn’t been updated to understand WebTransport addresses, it won’t be able to parse a WebTransport multiaddress, and won’t be able to forward it to other (WebTransport-enabled) nodes.

Related: https://github.com/multiformats/multiaddr/issues/155

Minor problems:

- `Multiaddr` is an interface, and therefore (almost always) allocated on the heap. This creates a lot of GC pressure.
- Methods are split between packet-level functions (e.g. `SplitLast`) and methods on the interface (e.g. `ValueForProtocol`). There’s no (apparent) logic behind this, making use of the API confusing
- The methods weren’t designed with allocations in mind. One of them (especially `Protocols`) is the major source of allocations in a Kubo node

Tasks
- [x] https://github.com/multiformats/go-multiaddr/pull/261
- [x] https://github.com/multiformats/go-multiaddr/pull/271
- [x] https://github.com/multiformats/go-multiaddr/pull/272
- [x] https://github.com/multiformats/go-multiaddr/pull/273
- [x] https://github.com/multiformats/go-multiaddr/pull/274
- [x] https://github.com/multiformats/go-multiaddr/pull/270
- [x] https://github.com/multiformats/go-multiaddr/pull/265
- [x] https://github.com/multiformats/go-multiaddr/pull/269
- [x] https://github.com/multiformats/go-multiaddr/pull/263

MarcoPolo:
Another annoyance I ran into: `Component` doesn't implement the Multiaddr interface. So if you collect a bunch into a slice: `[]Components` there's no way to join them all. I'm not sure if we even need the component in the new API, just wanted to mention it as another pain point.

PR #200, opened 2023-06-03
reduce allocations in net.ResolveUnspecifiedAddress

sukunrt:
On the benchmark it reduces allocations from 10 allocs/op to 5 allocs/op. The benefits are greater if you have multiple interface addresses.

### DIFF
--- a/net/resolve.go
+++ b/net/resolve.go
@@ -7,26 +7,26 @@ import (
 )
 
 // ResolveUnspecifiedAddress expands an unspecified ip addresses (/ip4/0.0.0.0, /ip6/::) to
-// use the known local interfaces. If ifaceAddr is nil, we request interface addresses
-// from the network stack. (this is so you can provide a cached value if resolving many addrs)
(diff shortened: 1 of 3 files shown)


net/resolve.go
```diff
@@ -7,26 +7,31 @@ import (
...
-	if !IsIPUnspecified(split[0]) {
+	if !IsIPUnspecified(first) {
 		return []ma.Multiaddr{resolve}, nil
 	}
 
 	out := make([]ma.Multiaddr, 0, len(ifaceAddrs))
 	for _, ia := range ifaceAddrs {
 		// must match the first protocol to be resolve.
-		if ia.Protocols()[0].Code != resolve.Protocols()[0].Code {
+		match := false
```
marten-seemann, line 26: This would be solved by #198.

sukunrt: That's fair. It's still inefficient and it's very difficult to optimise this beyond the method in the PR without doing #198.

PR #202, opened 2023-06-29
manet: add function to test if address is NAT64 IPv4 converted IPv6 address

sukunrt:


net/ip_test.go
```diff
@@ -0,0 +1,44 @@
...
+	}{
+		{
+			addr:          ma.StringCast("/ip4/1.2.3.4/tcp/1234"),
+			want:          false,
+			failureReason: "ip4 addresses should return false",
+		},
+		{
+			addr:          ma.StringCast("/ip6/1::4/tcp/1234"),
+			want:          false,
+			failureReason: "ip6 addresses doesn't have wellknown prefix",
```
marten-seemann, line 24:
```suggestion
			failureReason: "ip6 addresses doesn't have well-known prefix",
```

marten-seemann (approved):

net/ip_test.go
sukunrt: fixed

PR #203, opened 2023-06-29
add a Unique function

marten-seemann:
This is the function we optimized in https://github.com/libp2p/go-libp2p/pull/2395. The plan is to remove this function from go-libp2p.

### DIFF
--- a/multiaddr.go
+++ b/multiaddr.go
@@ -213,9 +213,9 @@ func Contains(addrs []Multiaddr, addr Multiaddr) bool {
 	return false
 }
 
-// Deduplicate deduplicates addresses in place, leave only unique addresses.
+// Unique deduplicates addresses in place, leave only unique addresses.
 // It doesn't allocate.
-func Deduplicate(addrs []Multiaddr) []Multiaddr {
+func Unique(addrs []Multiaddr) []Multiaddr {
 	if len(addrs) == 0 {
 		return addrs
 	}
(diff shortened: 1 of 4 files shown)


multiaddr.go
```diff
@@ -210,3 +212,24 @@ func Contains(addrs []Multiaddr, addr Multiaddr) bool {
 	}
 	return false
 }
+
+// Deduplicate deduplicates addresses in place, leave only unique addresses.
+// It doesn't allocate.
+func Deduplicate(addrs []Multiaddr) []Multiaddr {
```
marten-seemann, line 218: Maybe better: Unique? Any preferences?

MarcoPolo: Preference for Unique

MarcoPolo (commented):
I think there should be a Unique and a UniqueSorted. UniqueSorted being the version that works if the list is already sorted.

multiaddr.go
marten-seemann: Updated to Unique.

marten-seemann:
> I think there should be a Unique and a UniqueSorted. UniqueSorted being the version that works if the list is already sorted.

Updated to Unique. We can add a UniqueSorted later, if the need arises.

sukunrt (approved):

sukunrt:
Unique is nice.

PR #207, opened 2023-07-26
fix: update /x/exp package and fix sorting

paralin:
Update /x/exp to match the go 1.21 "slices" package.

The signature of slices.SortFunc changed to now require the function return an integer instead of a boolean for sorting. Fix the call to slices.SortFunc.

This fixes the build on Go 1.21 when downstream projects import a newer version of /x/exp than the current version of multiaddr.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -8,7 +8,7 @@ require (
 	github.com/multiformats/go-multihash v0.0.14
 	github.com/multiformats/go-varint v0.0.6
 	github.com/stretchr/testify v1.7.0
-	golang.org/x/exp v0.0.0-20230626212559-97b1e661b5df
+	golang.org/x/exp v0.0.0-20230725093048-515e97ebf090
 )
 
 require (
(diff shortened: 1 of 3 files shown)


go.mod
```diff
@@ -8,13 +8,14 @@ require (
...
-	golang.org/x/exp v0.0.0-20230626212559-97b1e661b5df
+	golang.org/x/exp v0.0.0-20230725093048-515e97ebf090
 )
 
 require (
 	github.com/davecgh/go-spew v1.1.0 // indirect
+	github.com/klauspost/cpuid/v2 v2.2.3 // indirect
 	github.com/minio/blake2b-simd v0.0.0-20160723061019-3f5f724cb5b1 // indirect
-	github.com/minio/sha256-simd v0.1.1-0.20190913151208-6de447530771 // indirect
+	github.com/minio/sha256-simd v1.0.1 // indirect
```
marten-seemann, line 18: Any reason to update this dependency? (Same for cupid as well)

paralin: I believe those were updated by go mod tidy as dependencies of the sha package.

paralin: Ah: I updated the sha package to fix a test failure due to a bug in the amd64 simd instructions. The newer version works and the older fails with a fatal error I've never seen before about the stack failing to unwind (or something like that - off the top of my head.)

marten-seemann: Can you please open a separate PR for any other dependency updates? Trying to keep the changes here minimal.

muXxer: Would be cool to fix this `golang.org/x/exp` package problem soon... because it breaks my app :(

muXxer: Or would it make sense to use the sort.Slice function instead until go 1.21 is released, even if that allocates memory compared to slices.Sort?

marten-seemann: Allocating is not an option, this function is getting called a lot. Should be easy enough to clean up this PR though. @paralin, are you planning to make these changes?

jorgemmsilva: +1 please merge this, we cannot update `/x/exp` package until this is fixed.

paralin: Fixed.

paralin:
@marten-seemann Fixed - split into two PRs. Confirmed that the other PR is necessary to fix a go1.21 test failure: #212 but it's unrelated to this PR.

marten-seemann:
Fixed by #210.

PR #210, opened 2023-08-02
update golang.org/x/exp, fixing slice sorting to match standard library

cce:
This updates the dependency on the slices package from golang.org/x/exp used here to match the upcoming Go 1.21 standard library API, which uses a different comparison function.

https://github.com/golang/exp/commit/302865e7556b4ae5de27248ce625d443ef4ad3ed
https://github.com/golang/go/issues/61374


marten-seemann:
@cce Why are you closing this PR?

marten-seemann (approved):

PR #220, opened 2023-09-07
net: consider dns addresses as public

sukunrt:
This resolves the discussion here: https://github.com/libp2p/go-libp2p/pull/2553#pullrequestreview-1612350763

For some users this might cause unexpected behaviour e.g. [here](https://github.com/kubeedge/edgemesh/blob/3d3c9193b58943171031f4dabdd2f37b8d1cb972/pkg/tunnel/tunnel.go#L53) it looks like the user does want this function to return true only for IP addresses. I can imagine there are some others who want this.

Most of the [cases](https://github.com/search?q=manet.IsPublicAddr&type=code) I eyeballed are not affected by this change and in some cases are positively impacted as they do expect dns addresses to pass this check.

Ideally it'd have been nicer if the existing method was IsPublicIPAddr and so we could just add IsPublicAddr with this behaviour.

The suffix check is not too expensive. The method takes 90-100ns with IP addresses. It takes 125ns with dns addresses on my m1 macbook.

### DIFF
--- a/net/private.go
+++ b/net/private.go
@@ -113,7 +113,7 @@ func IsPublicAddr(a ma.Multiaddr) bool {
 			for _, ud := range unResolvableDomains {
 				if strings.HasSuffix(dnsAddr, ud) {
 					isPublic = false
-					break
+					return false
 				}
 			}
 			for _, pd := range privateUseDomains {
--- a/net/private.go
+++ b/net/private.go
@@ -2,6 +2,7 @@ package manet
 
 import (
 	"net"
+	"strings"
 
 	ma "github.com/multiformats/go-multiaddr"
 )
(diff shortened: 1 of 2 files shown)


sukunrt:
I've fixed the list of reserved names. 

I'm rejecting all reverse dns lookup addresses because they work only with PTR DNS queries.

net/private.go
```diff
@@ -78,6 +107,21 @@ func IsPublicAddr(a ma.Multiaddr) bool {
 		case ma.P_IP6:
 			ip := net.IP(c.RawValue())
 			isPublic = !inAddrRange(ip, Private6) && !inAddrRange(ip, Unroutable6)
+		case ma.P_DNS, ma.P_DNS4, ma.P_DNS6, ma.P_DNSADDR:
+			dnsAddr := c.Value()
+			isPublic = true
+			for _, ud := range unResolvableDomains {
+				if strings.HasSuffix(dnsAddr, ud) {
+					isPublic = false
+					break
```
marten-seemann, line 116: Why not return, otherwise you'll be entering the next loop.

marten-seemann (commented):
Lists look reasonable and roughly what I would've expected.

Can you add a comment where you took these from?

net/private.go
sukunrt: that's better. fixed.

sukunrt:
The individual entries have the corresponding RFC except reverse lookup domains. Should I mention that this is a cut down version of special use domains from wikipedia?

marten-seemann:
Yes, it's fine to cite Wikipedia here. I just want to make sure that if we need to come back to this at some point in the future, we can figure out where this originated.

marten-seemann (approved):

PR #221, opened 2023-09-16
net: consider /dns/localhost as private address

sukunrt:
This fixes IsPrivateAddr check to correctly classify localhost dns addresses as private. 

This fixes sub domain check to correctly classify equal domains as sub domains. So /dns/localhost/ is classified as non public. This is similar to https://pkg.go.dev/github.com/miekg/dns#IsSubDomain. I didn't want to import a library for this so I'm using a custom implementation.

### DIFF
--- a/net/private.go
+++ b/net/private.go
@@ -113,18 +113,18 @@ func IsPublicAddr(a ma.Multiaddr) bool {
 		case ma.P_DNS, ma.P_DNS4, ma.P_DNS6, ma.P_DNSADDR:
 			dnsAddr := c.Value()
 			isPublic = true
-			if isSubDomain(dnsAddr, localHostDomain) {
+			if isSubdomain(dnsAddr, localHostDomain) {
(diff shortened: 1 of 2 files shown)


net/private.go
```diff
@@ -130,6 +135,13 @@ func IsPublicAddr(a ma.Multiaddr) bool {
 	return isPublic
 }
 
+// isSubDomain checks if child is sub domain of parent. It also returns true if child and parent are
+// the same domain.
+// Parent must have a "." prefix.
+func isSubDomain(child, parent string) bool {
```
marten-seemann, line 141: Nit: I believe the more correct capitalization would be `isSubdomain`.

marten-seemann (approved):

net/private.go
sukunrt: indeed.

marten-seemann (approved):

PR #226, opened 2024-01-10
v0.12.1 bump

rvagg:


sukunrt (approved):

rvagg:
not sure why it's suggesting v0.13.0, I don't think that's justified; but the indirect dep bumps should satisfy dependabot security pings though

PR #228, opened 2024-01-23
Small fixes to parsing ports

0xTylerHolmes:
### What type of PR is this?

Bug fix 

This PR addresses some inconsistencies in the parsing of ports in various formats.

1. No longer allows negative integer port values to be silently converted to uints 
2. Remove incorrect error string for max uint16 values. (Previously the wrong value for a MAX_UINT16 was logged)
3. Replace AtoI and ItoA strcov calls with ParseUint and FormatUint
4. Add test for negative udp/tcp ports

### Potential Changes

I elected to replace the AtoI and ItoA with the parse/format uints with base16. These require less checks at the cost of less robust errors. If this seems like too big of a trade off I can reformat it back to the old way while still maintaining the bug fixes.

### DIFF
--- a/multiaddr_test.go
+++ b/multiaddr_test.go
@@ -83,6 +83,7 @@ func TestConstructFails(t *testing.T) {
 		"/ip4/127.0.0.1/p2p/tcp",
 		"/unix",
 		"/ip4/1.2.3.4/tcp/80/unix",
+		"/ip4/1.2.3.4/tcp/-1/unix",
 		"/ip4/127.0.0.1/tcp/9090/http/p2p-webcrt-direct",
 		"/",
 		"",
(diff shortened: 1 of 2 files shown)


multiaddr_test.go
```diff
@@ -83,6 +83,7 @@ func TestConstructFails(t *testing.T) {
 		"/ip4/127.0.0.1/p2p/tcp",
 		"/unix",
 		"/ip4/1.2.3.4/tcp/80/unix",
+		"/ip4/1.2.3.4/tcp/-1/unix",
```
Jorropo, line 86:
```suggestion
		"/ip4/1.2.3.4/tcp/-1",
```

Jorropo (changes requested):
That a great fix thx, please update the test so it only test the invalid port (not invalid port and invalid unix component)

Jorropo (approved):
LGTM thx for this

PR #229, opened 2024-01-23
Redundant nil checks

0xTylerHolmes:
### What type of PR is this?

Robustness + typo

Disclaimer: I have not found cases where proper library usage could trigger a panic on these functions. 

This PR does two things.

1. places additional nil checks on some of the SplitFirst calls. It adds a negligible performance cost to protect against improper usage in callers as well as protecting against panics caused by bugs in multiaddr parsing. 
2. fix typo in comments

### DIFF
--- a/net/ip.go
+++ b/net/ip.go
@@ -114,6 +114,9 @@ func zoneless(m ma.Multiaddr) ma.Multiaddr {
 			return nil
 		}
 		tailhead, _ := ma.SplitFirst(tail)
+		if tailhead == nil {
+			return nil
+		}
 		if tailhead.Protocol().Code != ma.P_IP6 {
 			return nil
 		}
--- a/net/resolve.go
+++ b/net/resolve.go
@@ -12,6 +12,9 @@ import (
 func ResolveUnspecifiedAddress(resolve ma.Multiaddr, ifaceAddrs []ma.Multiaddr) ([]ma.Multiaddr, error) {
 	// split address into its components
 	first, rest := ma.SplitFirst(resolve)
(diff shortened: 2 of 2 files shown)


net/ip.go
```diff
@@ -114,6 +114,9 @@ func zoneless(m ma.Multiaddr) ma.Multiaddr {
 			return nil
 		}
 		tailhead, _ := ma.SplitFirst(tail)
+		if tailhead == nil {
+			return nil
+		}
```
sukunrt, line 119: if tail is not nil tailhead has to be nonnil, right?

net/resolve.go
```diff
@@ -12,6 +12,9 @@ import (
 func ResolveUnspecifiedAddress(resolve ma.Multiaddr, ifaceAddrs []ma.Multiaddr) ([]ma.Multiaddr, error) {
 	// split address into its components
 	first, rest := ma.SplitFirst(resolve)
+	if first == nil {
+		return nil, fmt.Errorf("invalid address components: %s", resolve)
+	}
```
sukunrt, line 17: I think we should move the checks to the top of the function. 

```
if resolve == nil {
	return nil, fmt.Error("invalid arg resolve: <nil>")
}
for _, a := range ifaceAddrs {
    if a == nil {
    	return nil, fmt.Error("<nil> iface addrs")
    }
}
```

sukunrt:
Why did you close this? I'm interested in getting this fix in if you still want to: https://github.com/multiformats/go-multiaddr/pull/229#discussion_r1465881697

PR #231, opened 2024-01-24
return err on empty multiaddr

0xTylerHolmes:
### What type of PR is this?

Bug fix

This PR addresses an issue where the validation for the string and bytes constructors behave differently. validateBytes() checks that the multiaddr is not empty and returns an error if so. The string constructor does not which can lead an empty multiaddress with no error that will cause a panic when operated on.

### DIFF
--- a/multiaddr.go
+++ b/multiaddr.go
@@ -27,6 +27,10 @@ func NewMultiaddr(s string) (a Multiaddr, err error) {
 	if err != nil {
 		return nil, err
 	}
+	if len(b) == 0 {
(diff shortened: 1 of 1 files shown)


sukunrt:
This is the behaviour already: https://github.com/multiformats/go-multiaddr/blob/master/codec.go#L25

0xTylerHolmes:
Oh you're right. Sorry about that, I had some issues in a local version of the repo I was doing some testing on. Must've got turned around. Closing this.

PR #235, opened 2024-02-12
net: restrict unicast ip6 public address space

sukunrt:
Global Unicast IPv6 addresses only belong to prefix 2000::/3. This change also classifies NAT64 prefixes as Public Addresses as they may reference public IPv4 addresses.

### DIFF
--- a/net/private_test.go
+++ b/net/private_test.go
@@ -53,6 +53,21 @@ func TestIsPublicAddr(t *testing.T) {
 			isPublic:  false,
 			isPrivate: true,
 		},
+		{
+			addr:      ma.StringCast("/ip6/2400::1/tcp/10"),
+			isPublic:  true,
+			isPrivate: false,
+		},
+		{
+			addr:      ma.StringCast("/ip6/2001:db8::42/tcp/10"),
+			isPublic:  false,
+			isPrivate: false,
+		},
+		{
+			addr:      ma.StringCast("/ip6/64:ff9b::1.1.1.1/tcp/10"),
+			isPublic:  true,
+			isPrivate: false,
+		},
 	}
 	for i, tt := range tests {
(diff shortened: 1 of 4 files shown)


net/private_test.go
```diff
@@ -53,6 +53,21 @@ func TestIsPublicAddr(t *testing.T) {
...
+		{
+			addr:      ma.StringCast("/ip6/2001:db8::42/tcp/10"),
+			isPublic:  false,
+			isPrivate: false,
+		},
+		{
+			addr:      ma.StringCast("/ip6/64:ff9b::1.1.1.1/tcp/10"),
+			isPublic:  true,
+			isPrivate: false,
+		},
```
Jorropo, line 70: Shouldn't it fallback to ipv4 isprivate logic ?
I mean if I use NAT64 to encode `192.168.0.1` then it should report `isPublic: false, isPrivate: true,`.

sukunrt: Pedantically that's illegal, the well known prefix should only be used with public addresses and the other prefix leaves it to the users how they convert between ip6 and ip4. 
I think this is fine to add though. I'll update.

Jorropo: If some RFC defines this as illegal that ok, no need to implement weird edge cases.

sukunrt: https://datatracker.ietf.org/doc/html/rfc6052#section-3.1

```
The Well-Known Prefix MUST NOT be used to represent non-global IPv4
   addresses, such as those defined in [[RFC1918](https://datatracker.ietf.org/doc/html/rfc1918)] or listed in Section 3
   of [RFC5735].
```

Jorropo: thx

MarcoPolo (approved):

PR #242, opened 2024-04-07
feat: add tinygo support

paralin:
Add a separate build-tagged file for tinygo and not-tinygo (other).

Stub functions not available in tinygo.

Fixes build with:

tinygo build -tags "tinygo purego" -target wasm -no-debug -opt=2 ./


Stebalien:
It looks like the upstream commit is merged. Is this still needed?

paralin:
I will check, thanks

PR #246, opened 2024-06-05
Add support for http-path

MarcoPolo:
Context here: https://github.com/multiformats/multiaddr/pull/164

### DIFF
--- a/multiaddr_test.go
+++ b/multiaddr_test.go
@@ -627,6 +627,7 @@ func TestRoundTrip(t *testing.T) {
 		"/ip4/127.0.0.1/udp/1234/quic-v1/webtransport/certhash/uEiDDq4_xNyDorZBH3TlGazyJdOWSwvo4PUo5YHFMrvDE8g",
 		"/p2p/QmbHVEEepCi7rn7VL7Exxpd2Ci9NNB6ifvqwhsrbRMgQFP",
 		"/p2p/QmbHVEEepCi7rn7VL7Exxpd2Ci9NNB6ifvqwhsrbRMgQFP/unix/a/b/c",
+		"/http-path/tmp%2Fbar",
 	} {
 		ma, err := NewMultiaddr(s)
 		if err != nil {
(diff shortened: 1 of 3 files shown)


multiaddr_test.go
```diff
@@ -923,3 +924,36 @@ func TestDNS(t *testing.T) {
...
+		badAddr := "/http-path/thisIsMissingAfullBytes%f"
+		_, err := NewMultiaddr(badAddr)
+		require.Error(t, err)
+	})
+
+	t.Run("round trip", func(t *testing.T) {
+		cases := []string{
+			"/http-path/tmp%2Fbar",
+			"/http-path/tmp%2Fbar%2Fbaz",
+			"/http-path/foo",
```
sukunrt, line 939: Can you add a case with the complete multiaddr with ipv4 + p2p + http + http-path or just ipv4 + p2p + http-path?

multiaddr_test.go
```diff
@@ -923,3 +924,36 @@ func TestDNS(t *testing.T) {
 		t.Fatal("expected equality")
 	}
 }
+
+func TestHTTPPath(t *testing.T) {
+	t.Run("bad addr", func(t *testing.T) {
+		badAddr := "/http-path/thisIsMissingAfullBytes%f"
+		_, err := NewMultiaddr(badAddr)
```
sukunrt, line 931: Can we add a test case for 
`/http-path/X/Y` If I understand it correctly for the returned multiaddr `addr.ValueForProtocol(ma.P_HTTP_PATH)` should have the value `X`

sukunrt (commented):
Can we add a few examples of the path to the fuzzer here: https://github.com/multiformats/go-multiaddr/blob/e0315b9835454ee19c4e3ac222b1b9215da22bda/multiaddr_test.go#L107

MarcoPolo:
good call on the fuzzing. I found a surprising behavior where if a variable sized component has a zero-len value it will not put anything into the multiaddr string. This means that the string `/http-path/http` is ambigous since it's not clear if it's `http-path` with no value and an `http` component, or an `http-path` with the value "http". 

To disambiguate, I've made it a requirement for there to be at least some value in the `http-path` component. Which seems to be the same thing that other length prefixed sized components do (e.g. dns, p2p). This library should probably make it possible to represent components with a value that has len 0 (e.g. `/http-path//http`), but nothing currently needs this, so likely not a priority.

sukunrt (approved):

PR #250, opened 2024-07-01
Make it safe to roundtrip Split... and Join

MarcoPolo:


multiaddr_test.go
```diff
@@ -974,3 +974,43 @@ func TestHTTPPath(t *testing.T) {
...
+		joined := Join(first, rest)
+		require.Equal(t, addr, joined, "SplitFirst and Join should round-trip")
+
+		// Test SplitLast
+		rest, last := SplitLast(addr)
+		joined = Join(rest, last)
+		require.Equal(t, addr, joined, "SplitLast and Join should round-trip")
+
+		p := addr.Protocols()
+		if err != nil || len(p) == 0 {
```
sukunrt, line 1000:
```suggestion
		if len(p) == 0 {
```

There's no error.

multiaddr_test.go
```diff
@@ -974,3 +974,43 @@ func TestHTTPPath(t *testing.T) {
...
+			t.Skip() // Skip if no components or error
+		}
+
+		selectedProto := p[int(splitOnProto)%len(p)]
```
sukunrt, line 1004: You can split on all protos in a loop.

util.go
```diff
@@ -28,12 +28,18 @@ func Join(ms ...Multiaddr) Multiaddr {
...
 
 	bidx := 0
 	b := make([]byte, length)
 	for _, mb := range ms {
+		if mb == nil {
+			continue
+		}
 		bidx += copy(b[bidx:], mb.Bytes())
 	}
 	return &multiaddr{bytes: b}
```
sukunrt, line 45: This will return a multiaddr with empty bytes when passed all nils. Should we have a special check for that? This seems okay, but I'm not sure how this empty multiaddress will behave everywhere else.

sukunrt (approved):

util.go
sukunrt: Should we Fuzz test calling `String()`, `Protocols()` etc on the results of `SplitFirst` and `SplitLast?`

sukunrt (approved):

PR #255, opened 2024-09-29
fix: `permission denied` appears on Android11+

wlynxg:
fix: 
- https://github.com/libp2p/go-libp2p/issues/2986
- https://github.com/ipfs/kubo/issues/9513

related:
- https://github.com/golang/go/issues/40569
- https://github.com/golang/go/pull/61089

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -20,6 +20,7 @@ require (
 	github.com/multiformats/go-base36 v0.2.0 // indirect
 	github.com/pmezard/go-difflib v1.0.0 // indirect
 	github.com/spaolacci/murmur3 v1.1.0 // indirect
+	github.com/wlynxg/anet v0.0.4 // indirect
 	golang.org/x/crypto v0.18.0 // indirect
 	golang.org/x/sys v0.16.0 // indirect
 	gopkg.in/yaml.v3 v3.0.0 // indirect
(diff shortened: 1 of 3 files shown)


marten-seemann:
Pulling in a fork of the `net` package seems like the worst way possible to solve the problem.

wlynxg:
If go can merge pr, it would be the best way. However, due to the variability of Android versions, merging into the main branch seems very difficult. 
Therefore, although using a third-party library to obtain the list of network card interfaces is not the best way, it has been the solution on the Android side for a long time.

homier:
@marten-seemann for now it's the only way. Go maintainers do not really do anything about the issue, the PR is stuck for several months in a row without any progress, while a lot of folks that can't use anything related to multiaddr on Android without compiling Go toolchain with patch.

fyi: pion-webrtc also took this approach with `anet` (https://github.com/pion/transport/commit/3ba6182e023088a6242d24493b39c41b182227b2)

gbraad:
I understand the sentiment of not wanting an abstraction layer to fix the issue, but without putting pressure on the upstream Golang issue, this will remain in this limbo.

MarcoPolo:
I don't love the solutions here.

Here's another solution I don't love, but maybe less bad than the others:
1. Fork net ourselves (or vendor it with patches)
2. Add a build tag to only include the fork on android builds.

marten-seemann:
`InterfaceMultiaddrs` is a helper function, which one could argue doesn't belong in this package in the first place, or at least take a `[]net.Addr` as input. It should be absolutely trivial for users to copy-paste it. I don't see any reason whatsoever to pull in a 3rd-party dependency.

Stebalien:
IMO, the right solution is a package that exposes _just_ a fixed `InterfaceAddrs` function. Overriding the entire `net` package is asking for trouble. That's also something that can be easily vendor.

PR #261, opened 2025-01-11
refactor!: make the API harder to misuse

MarcoPolo:
This seeks to refactor the codebase to make it much harder to hit nil pointer dereference panics.

This takes a different approach to how we've treated multiaddrs in the past. Instead of attempting to make them a general and performant datastructure, we focus on treating them as just an encoding scheme. Users of multiaddrs are expected to parse the multiaddr into some struct that is suitable for their use case, and use the multiaddr form when interoperating. By treating Multiaddrs as just an encoding scheme we can make a number of simplifications in the codebase. Specifically this PR does the following:

- Removes the `Multiaddr`interface.
- Multiaddr is now a concrete type of `[]Component`
- `Component`is now a public type, and users can use `Component`s directly
- Components no longer implement the `Multiaddr`interface as there is none.

## Background

This library has had multiple issues related to `Multiaddr` being an interface. Many methods use and return `nil` as the zero value, which behaves poorly when the user forgets to do a nil check on every returned value and attempts to call a method on the nil pointer. For example, using `Split` to split a Multiaddr and then using `Join` to rebuild the original Multiaddr historically would panic in case one side of the split was nil. Using an interface also leads to incorrect usages of `==` to check if two Multiaddrs were equal (would only work for pointer equality) and incorrectly using `Multiaddr` as a key for a map.

Using an interface is typically done to provide a consistent API surface for multiple implementing types. In practice however, the Multiaddr interface was only implemented for `multiaddr` and `component` (with arguably some awkwardness when using a component as a Multiaddr).

The better approach is to use a concrete type for a Multiaddr. This lets pointer receiver methods work even if the pointer is nil, since the compiler already knows which function to call. Most methods now take a value rather than a pointer which avoids the issue of a nil pointer dereference completely.

## Migration

Refer to ./v015-MIGRATION.md for breaking changes and migration tips


multiaddr.go
```diff
@@ -38,131 +51,149 @@ func NewMultiaddrBytes(b []byte) (a Multiaddr, err error) {
...
+		if cmp := m[i].Compare(o[i]); cmp != 0 {
+			return cmp
+		}
+	}
+	if len(m) < len(o) {
+		return -1
+	} else if len(m) > len(o) {
+		return 1
+	}
+	return 0
```
sukunrt, line 91: NIT: `return cmp.Compare(len(m), len(o))`

multiaddr.go
```diff
@@ -38,131 +51,149 @@ func NewMultiaddrBytes(b []byte) (a Multiaddr, err error) {
...
 }
 
 // Bytes returns the []byte representation of this Multiaddr
-//
-// Do not modify the returned buffer, it may be shared.
-func (m *multiaddr) Bytes() []byte {
-	return m.bytes
+func (m Multiaddr) Bytes() []byte {
+	size := 0
```
sukunrt, line 96: This is more expensive than it used to be. Worth calling out in release notes.

util.go
```diff
@@ -70,135 +62,61 @@ func StringCast(s string) Multiaddr {
 }
 
 // SplitFirst returns the first component and the rest of the multiaddr.
-func SplitFirst(m Multiaddr) (*Component, Multiaddr) {
```
sukunrt, line 73: Maybe it's better to return `(Multiaddr, Multiaddr)`
The first multiaddr will be `[]Component{first}`


With the current API, most users will need to do `component.AsMultiaddr` here. 

This does break users who were expecting a component, but I believe that's the less popular usecase.

MarcoPolo: I made this change and backed it out. I think a lot of usages was to get the component out, not a Multiaddr.

For example this is a pretty common pattern:
```go
transport, p2ppart := ma.SplitLast(m)
if p2ppart == nil || p2ppart.Protocol().Code != ma.P_P2P {
    return m, ""
}
```

If we return a Multiaddr type, you can't use the `Protocol()` method like you were able to before.

This pattern mostly still works after this refactor, you just have to check `p2ppart.Empty()` instead of `== nil`

component.go
```diff
@@ -12,20 +11,42 @@ import (
 
 // Component is a single multiaddr Component.
 type Component struct {
-	bytes    []byte
+	bytes    string // Uses the string type to ensure immutability.
 	protocol Protocol
```
sukunrt, line 15: We can also just keep a pointer to the `Protocol` and make this struct more compact.

component.go
```diff
@@ -173,14 +198,33 @@ func newComponent(protocol Protocol, bvalue []byte) *Component {
...
+		return Component{}, err
+
+	}
+	if c.protocol.Transcoder != nil {
+		err = c.protocol.Transcoder.ValidateBytes([]byte(c.bytes[c.offset:]))
+		if err != nil {
+			return Component{}, err
+		}
 	}
+	return c, nil
```
sukunrt, line 229: NIT

I'd move this logic inside `newComponent` and ensure that all components are created by `newComponent`.

Currently there are 4 places in code where we need to do `validateComponent(component{...})` and possibly all of them can be changed to `newComponent(...)`

interface.go
```diff
@@ -1,63 +0,0 @@
...
-Learn more here: https://github.com/multiformats/multiaddr
-
-Multiaddrs have both a binary and string representation.
-
-	import ma "github.com/multiformats/go-multiaddr"
-
-	addr, err := ma.NewMultiaddr("/ip4/1.2.3.4/tcp/80")
-	// err non-nil when parsing failed.
-*/
-type Multiaddr interface {
```
sukunrt, line 20: :tada:

matest/matest.go
```diff
@@ -0,0 +1,80 @@
...
+	"github.com/multiformats/go-multiaddr"
+)
+
+type TestingT interface {
+	Errorf(format string, args ...interface{})
+}
+
+type tHelper interface {
+	Helper()
+}
```
sukunrt, line 16: Does `testing.TB` work here?

matest/matest.go
```diff
@@ -0,0 +1,80 @@
...
+type MultiaddrMatcher struct {
+	multiaddr.Multiaddr
+}
+
+func (m MultiaddrMatcher) Matches(x interface{}) bool {
+	if m2, ok := x.(multiaddr.Multiaddr); ok {
+		return m.Equal(m2)
+	}
+	return false
+}
```
sukunrt, line 80: Why do we need this?

multiaddr.go
```diff
@@ -38,131 +52,144 @@ func NewMultiaddrBytes(b []byte) (a Multiaddr, err error) {
...
-	if right == nil {
+func (m Multiaddr) EncapsulateC(c Component) Multiaddr {
+	if c.Empty() {
 		return m
 	}
+	out := make([]Component, 0, len(m)+1)
+	out = append(out, m...)
+	out = append(out, c)
+	return out
+}
```
sukunrt, line 188: As the multiaddr is immutable, can we just appent to the existing slice?

sukunrt (approved):
This looks great. :heart:   

Left some comments; I don't feel strongly about them.

matest/matest.go
MarcoPolo: No. We want to support using this with assert's `EventuallyWithT` which gives us a type (`*assert.CollectT`) which does not implement `testing.TB`. The assert library does a similar thing [here](https://github.com/stretchr/testify/blob/master/mock/mock.go#L1200-L1202).

matest/matest.go
MarcoPolo: You're right, a comment would be helpful here.

https://pkg.go.dev/github.com/golang/mock/gomock#Matcher

This lets us match against other multiaddrs in our Go mock test. [Example](https://github.com/libp2p/go-libp2p/blob/b434cb255a407390b827b98ae781e1c5c30c71e8/p2p/test/transport/gating_test.go#L93)

multiaddr.go
MarcoPolo: Is the argument to remove this method or change this to append?

The caller is always free to use append directly. This method does a copy on purpose to avoid sharing the underlying slice. If the caller is okay with that they can `append` directly.

component.go
```diff
@@ -12,20 +11,42 @@ import (
 
 // Component is a single multiaddr Component.
 type Component struct {
-	bytes    []byte
+	bytes    string // Uses the string type to ensure immutability.
 	protocol Protocol
 	offset   int
```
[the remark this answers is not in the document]
MarcoPolo: renamed to valueStartIdx

component.go
MarcoPolo: I only see two usages of `validateComponent`:
1. In `newComponent`.
2. In `readComponent` (it happens twice in that method)

These two functions are a bit different. `newComponent` is expecting a creating a `Component` from parts, while `readComponent` is creating a Component from a byte array. Changing these methods to be the same is a bit awkward as `Component.bytes` is the full bytes representation of the component, so we would introduce a copy somewhere if we did so.

I think it's fine to leave as is.

PR #263, opened 2025-01-16
Megular Expressions

MarcoPolo:
A very simple regular expression matcher for Multiaddr components. Supports capturing values. Matches in linear time (no back tracking).

The core logic is about 100 LOC. The sugar to make it nicer to use is about another 100 LOC.

## Motivation

If we are going to treat Multiaddrs as encoding, then we need to make it more ergonomic to parse multiaddrs. Right now we have a lot of somewhat wrong manual parsers using `ForEach`. This should be able to replace those, make it cleaner, and most importantly make them obviously correct.

In draft while I try this out in go-libp2p.


meg/sugar.go
```diff
@@ -0,0 +1,130 @@
+package meg
+
+import (
+	"errors"
+)
+
```
sukunrt, line 6: I like the API. Some other things I'd like are: 

`Any` which matches any protocol

Match the pattern somewhere in the middle, not necessarily at the start.

`Any()*` which will match everything from that point onwards, so we can match IP, TCP, Port, Rest,

meg/meg.go
```diff
@@ -0,0 +1,126 @@
...
+
+	for _, c := range components {
+		if len(currentStates.states) == 0 {
+			return false, nil
+		}
+		for i, s := range currentStates.states {
+			if s.kind == matchCode && s.code == c.Code() {
+				cm := currentStates.captures[i]
+				if s.capture != nil {
+					cm[s.capture] = append(cm[s.capture], c.Value())
```
sukunrt, line 83: I think it's more powerful to capture the entire Component and not just Value. That way for some uses you don't have to move back from the string from to the component form.

meg/sugar.go
```diff
@@ -0,0 +1,130 @@
+package meg
```
sukunrt, line 1: This is a bit too slow: Consider these two implementations:

```
// IsWebRTCDirectMultiaddr returns whether addr is a /webrtc-direct multiaddr with the count of certhashes
// in addr
func IsWebRTCDirectMultiaddr(addr ma.Multiaddr) (bool, int) {
	var certHashes []string
	found, _ := addr.Match(
		meg.Or(
			meg.Val(ma.P_IP4),
			meg.Val(ma.P_IP6),
			meg.Val(ma.P_DNS),
		),
		meg.Val(ma.P_UDP),
		meg.Val(ma.P_WEBRTC_DIRECT),
		meg.CaptureZeroOrMore(ma.P_CERTHASH, &certHashes),
	)
	return found, len(certHashes)
}

// IsWebRTCDirectMultiaddrLoop returns whether addr is a /webrtc-direct multiaddr with the count of certhashes
// in addr
func IsWebRTCDirectMultiaddrLoop(addr ma.Multiaddr) (bool, int) {
	protos := [...]int{ma.P_IP4, ma.P_IP6, ma.P_DNS, ma.P_UDP, ma.P_WEBRTC_DIRECT}
	matchProtos := [...][]int{protos[:3], {protos[3]}, {protos[4]}}
	certHashCount := 0
	for i, c := range addr {
		if i >= len(matchProtos) {
			if c.Protocol().Code == ma.P_CERTHASH {
				certHashCount++
			} else {
				break
			}
		} else {
			found := false
			for _, proto := range matchProtos[i] {
				if c.Protocol().Code == proto {
					found = true
					break
				}
			}
			if !found {
				return false, 0
			}
		}
	}
	return true, certHashCount
}


```


```
func BenchmarkIsWebRTCDirectMultiaddr(b *testing.B) {
	addr := ma.StringCast("/ip4/1.2.3.4/udp/1234/webrtc-direct/")

	b.ResetTimer()
	for i := 0; i < b.N; i++ {
		isWebRTC, count := IsWebRTCDirectMultiaddr(addr)
		if !isWebRTC || count != 0 {
			b.Fatal("unexpected result")
		}
	}
}

func BenchmarkIsWebRTCDirectMultiaddrLoop(b *testing.B) {
	addr := ma.StringCast("/ip4/1.2.3.4/udp/1234/webrtc-direct/")

	b.ResetTimer()
	for i := 0; i < b.N; i++ {
		isWebRTC, count := IsWebRTCDirectMultiaddrLoop(addr)
		if !isWebRTC || count != 0 {
			b.Fatal("unexpected result")
		}
	}
}
```

On my machine: 
```
goos: linux
goarch: amd64
pkg: github.com/libp2p/go-libp2p/p2p/transport/webrtc
cpu: AMD Ryzen 7 7840U w/ Radeon  780M Graphics     
BenchmarkIsWebRTCDirectMultiaddr-16              1240466               943.5 ns/op           912 B/op         24 allocs/op
BenchmarkIsWebRTCDirectMultiaddrLoop-16         38312790                29.70 ns/op            0 B/op          0 allocs/op
PASS
ok      github.com/libp2p/go-libp2p/p2p/transport/webrtc        3.336s

```

I believe this diff mostly comes from all the closures used in implementing this which leads to a lot of allocations(Haven't debugged this too much). Can we find some less allocation heavy alternative to this structure? 

This isn't prohibitive the whole thing is only slightly slower than what it was before because previously the `ForEach` iteration was quite expensive.

sukunrt (commented):
This looks very useful to me modulo comments. See how nice the `IsWebRTCDirectMultiaddr` method is now vs compared to master. 

I think we should make this package Experimental / Alpha and start using this in go-libp2p. After some experience we can remove the Experimental tag, till then we can keep iterating on a satisfactory api.

meg/sugar.go
MarcoPolo: Thanks for the benchmark. I'm sure the overhead here is in the allocations. Still, this is better than I would have expected. All thanks to Go's GC and memory management. It's under 1 microsecond! 

I'll fiddle with this a bit and see if I can't improve it at least a little.

Note that the hand rolled iteration is slightly wrong. You should be returning false if you encounter a non cerhash component at then end rather than ignoring it.

util.go
```diff
@@ -122,3 +124,8 @@ func ForEach(m Multiaddr, cb func(c Component) bool) {
 		}
 	}
 }
+
+func (m Multiaddr) Match(p ...meg.Pattern) (bool, error) {
+	s := meg.PatternToMatchState(p...)
+	return meg.Match(s, m)
+}
```
sukunrt, line 131: Is there a reason to make this a method as opposed to:
 
`MatchPattern(m Multiaddr, p ...meg.Pattern)`

x/meg/sugar.go
```diff
@@ -0,0 +1,130 @@
...
+
+func Optional(s Pattern) Pattern {
+	return func(next *MatchState) *MatchState {
+		return &MatchState{
+			kind:      split,
+			next:      s(next),
+			nextSplit: next,
+		}
+	}
+}
```
sukunrt, line 130: Do we want an `Optional` or do we want a `MatchIfExists`? By `MatchIfExists` I mean if the component is there we *always* want to match it. While an Optional if I understand correctly, may match some part of the suffix too.

1. Might simplify implementation. We don't have to split. 
2. Might avoid ambiguity around did the optional match the prefix of the suffix etc.

x/meg/sugar.go
```diff
@@ -0,0 +1,130 @@
...
+	"errors"
+)
+
+type Pattern = func(next *MatchState) *MatchState
+
+func PatternToMatchState(states ...Pattern) *MatchState {
+	nextState := &MatchState{kind: done}
+	for i := len(states) - 1; i >= 0; i-- {
+		nextState = states[i](nextState)
+	}
```
sukunrt, line 13: Does this work with Split states at the beginning? The following example is failing for me:

```
func TestMatch(t *testing.T) {
	m := StringCast("/sni/ws")

	found, _ := m.Match(
		meg.Optional(meg.Val(P_SNI)),
		meg.Val(P_WS),
	)
	if !found {
		t.Fatal("failed to match", m)
	}
}
```

x/meg/meg.go
```diff
@@ -0,0 +1,126 @@
...
+		return arr
+	}
+	if c == nil {
+		c = make(captureMap)
+	}
+	s.generation = listGeneration
+	if s.kind == split {
+		arr = appendState(arr, s.next, c, listGeneration)
+		arr = appendState(arr, s.nextSplit, c.clone(), listGeneration)
+	} else {
```
sukunrt, line 121: What if the state added is also a split state? As I understand the Match loop expects non split states.

sukunrt (changes requested):

x/meg/sugar.go
MarcoPolo: What's the regex equivalent of MatchIfExists?

x/meg/sugar.go
MarcoPolo: This is fixes in the optimized branch, but this branch forgot to update the list generation after the initial appendStates. Here's the oneline fix:
```diff
diff --git a/x/meg/meg.go b/x/meg/meg.go
index 952111f..ba71795 100644
--- a/x/meg/meg.go
+++ b/x/meg/meg.go
@@ -74,6 +74,7 @@ func Match[S ~[]T, T Matchable](s *MatchState, components S) (bool, error) {
 	}
 
 	currentStates = appendState(currentStates, s, nil, listGeneration)
+	listGeneration++
 
 	for _, c := range components {
 		if len(currentStates.states) == 0 {

```

util.go
MarcoPolo: No reason. I think it's a bit more ergonomic for it to be a method, but I don't have a strong preference

x/meg/meg.go
MarcoPolo: I don't understand the comment. This is a recursive function, if `s.next` or `s.nextSplit` is a split it will be handled recursively until we have no splits.

util.go
sukunrt: I expect most users to not want to run this everytime. `	s := meg.PatternToMatchState(p...)`

sukunrt (approved):
Just one unaddresses comment: 
https://github.com/multiformats/go-multiaddr/pull/263#discussion_r1943467129

meg/meg.go
MarcoPolo: This is done in #269

util.go
MarcoPolo: I don't follow. They would just do:

```go
var m Multiaddr
// ...
m.Match(<some pattern>)
``` 

No use of `meg.PatternToMatchState` directly

meg/sugar.go
MarcoPolo: We have any now. Here's an example that captures a ipport and any*

```
found, err := m.Match(
    CaptureAddrPort(&network, &addrPort),
    meg.ZeroOrMore(meg.Any),
)
```

x/meg/sugar.go
MarcoPolo: > Implementation wise, if the next character doesn't match we still move the state ahead without moving the character index ahead.

This sounds a bit like Optional.

Overall, I think we do want Optional. For example, SNI is an optional field in many multiaddrs

x/meg/meg.go
```diff
@@ -0,0 +1,222 @@
...
+		currentStates, nextStates = nextStates, currentStates
+		nextStates.states = nextStates.states[:0]
+		nextStates.captures = nextStates.captures[:0]
+	}
+
+	for i, stateIndex := range currentStates.states {
+		s := states[stateIndex]
+		if s.codeOrKind == done {
+
+			// We found a complete path. Run the captures now
```
sukunrt, line 140: We need some notion of best path / longest path. 
```
func TestMeg(t *testing.T) {
	m := StringCast("/ip4/1.2.3.4/udp/8231/quic-v1/webtransport/certhash/b2uaraocy6yrdblb4sfptaddgimjmmpy/certhash/zQmbWTwYGcmdyK9CYfNBcfs9nhZs17a6FQ4Y8oea278xx41")

	var certhashes []string
	found, _ := m.Match(
		meg.ZeroOrMore(meg.Any),
		meg.CaptureOneOrMoreStringVals(P_CERTHASH, &certhashes),
	)
	if !found {
		t.Fatal("failed to match")
	}
	if len(certhashes) != 2 {
		t.Fatal("Didn't capture all certhashes")
	}

	{
		m, c := SplitLast(m)
		if c.Value() != certhashes[1] {
			t.Fatal("unexpected value. Expected", c.RawValue(), "but got", []byte(certhashes[1]))
		}
		_, c = SplitLast(m)
		if c.Value() != certhashes[0] {
			t.Fatal("unexpected value. Expected", c.RawValue(), "but got", []byte(certhashes[0]))
		}
	}
}

```
This test fails to match the certhashes

x/meg/sugar.go
```diff
@@ -0,0 +1,240 @@
...
+	"errors"
+	"fmt"
+	"strconv"
+	"strings"
+)
+
+// Pattern is essentially a curried MatchState.
+// Given the slice of current MatchStates and a handle (int index) to the next
+// MatchState, it returns a handle to the inserted MatchState.
+type Pattern = func(states *[]MatchState, nextIdx int) int
```
sukunrt, line 13: Would this be simpler if we returned both the modified MatchState slice and the nextIdx:

```
type Pattern = func(states []MatchState, nextIdx int) ([]MatchState, int)
```

x/meg/sugar.go
```diff
@@ -0,0 +1,240 @@
+package meg
+
+import (
+	"errors"
+	"fmt"
+	"strconv"
+	"strings"
+)
+
+// Pattern is essentially a curried MatchState.
```
sukunrt, line 10:
```suggestion
// Pattern is a curried MatchState.
```

x/meg/sugar.go
```diff
@@ -0,0 +1,240 @@
...
+
+		return splitIdx
+	}
+}
+
+func CaptureZeroOrMoreBytes(code int, vals *[][]byte) Pattern {
+	return CaptureZeroOrMoreWithF(code, captureManyBytes(vals))
+}
+
+func CaptureZeroOrMoreStringVals(code int, vals *[]string) Pattern {
```
sukunrt, line 203:
```suggestion
func CaptureZeroOrMoreStrings(code int, vals *[]string) Pattern {
```

x/meg/sugar.go
```diff
@@ -0,0 +1,240 @@
...
+	return CaptureZeroOrMoreWithF(code, captureManyStrings(vals))
+}
+
+func OneOrMore(code int) Pattern {
+	return CaptureOneOrMoreStringVals(code, nil)
+}
+
+func CaptureOneOrMoreStringVals(code int, vals *[]string) Pattern {
```
sukunrt, line 211: same comment here StringVals -> Strings

x/meg/meg.go
```diff
@@ -0,0 +1,222 @@
...
+
+// ListOfMatchable is anything list-like that contains Matchable items.
+// This allows us to convert a slice of []T as a []Matchable when *T implements Matchable.
+// In the future, this may not be required if Go generics allows us to say S ~[]T, and *T implements Matchable.
+type ListOfMatchable interface {
+	Get(i int) Matchable
+	Len() int
+}
+
+// Match returns whether the given Components match the Pattern defined in MatchState.
```
sukunrt, line 77:
```suggestion
// Match returns whether the given Components match the Matcher.
```

x/meg/sugar.go
```diff
@@ -0,0 +1,240 @@
...
+
+// Pattern is essentially a curried MatchState.
+// Given the slice of current MatchStates and a handle (int index) to the next
+// MatchState, it returns a handle to the inserted MatchState.
+type Pattern = func(states *[]MatchState, nextIdx int) int
+
+type Matcher struct {
+	states   []MatchState
+	startIdx int
+}
```
sukunrt, line 18: Can you add some documentation here? This is the type used by the `Match` function. 

Something simple like 
```
use PatternToMatcher to create these objects.  
```
is fine.

x/meg/meg.go
```diff
@@ -0,0 +1,222 @@
...
+)
+
+type stateKind = int
+
+const (
+	matchAny stateKind = (iota * -1) - 1
+	// done MUST be the last stateKind in this list. We use it to determine if a
+	// state is a split index.
+	done
+	// Anything that is less than done is a split index
```
sukunrt, line 19:
```suggestion
	// done MUST be the last stateKind in this list. 	
	// Anything that is less than done is a split index
	done

```

x/meg/meg.go
```diff
@@ -0,0 +1,222 @@
...
+			arr.states = append(arr.states, t.idx)
+			arr.captures = append(arr.captures, t.cap)
+		}
+	}
+	return arr
+}
+
+const splitIdxOffset = (-1 * (done - 1))
+
+func storeSplitIdx(codeOrKind int) int {
```
sukunrt, line 216:
```suggestion
func encodeSplitIdx(idx int) int {
```

x/meg/meg.go
```diff
@@ -0,0 +1,222 @@
...
+	return (codeOrKind + splitIdxOffset) * -1
+}
+
+func restoreSplitIdx(splitIdx int) int {
+	return (splitIdx * -1) - splitIdxOffset
+}
```
sukunrt, line 222:
```suggestion
func decodeSplitIdx(splitIdxStateKind int) int {
	return (splitIdxStateKind * -1) - splitIdxOffset
}
```

x/meg/meg.go
```diff
@@ -0,0 +1,222 @@
...
+		} else {
+			// Otherwise, it's a valid final state -- append it.
+			arr.states = append(arr.states, t.idx)
+			arr.captures = append(arr.captures, t.cap)
+		}
+	}
+	return arr
+}
+
+const splitIdxOffset = (-1 * (done - 1))
```
sukunrt, line 214:
```suggestion
const splitIdxOffset =  -(done - 1)
```

x/meg/meg.go
```diff
@@ -0,0 +1,222 @@
...
+	if s.codeOrKind == done {
+		return "done"
+	}
+	if s.codeOrKind < done {
+		return fmt.Sprintf("split{left: %d, right: %d}", s.next, restoreSplitIdx(s.codeOrKind))
+	}
+	return fmt.Sprintf("match{code: %d, next: %d}", s.codeOrKind, s.next)
+}
+
+type Matchable interface {
```
sukunrt, line 57: Do we need this because we don't want to depend on the multiaddr package? 
If so, we should add that as a comment somewhere.

x/meg/sugar.go
```diff
@@ -0,0 +1,240 @@
+package meg
```
sukunrt, line 1: Can you add a test for 
Cat, Or, and any other exposed Pattern that's not covered by tests.

sukunrt:
This looks good except for this comment: https://github.com/multiformats/go-multiaddr/pull/263#discussion_r2091991170

I'll approve once we resolve that.

x/meg/meg.go
MarcoPolo: Fixed in 002620b by making `Any` less greedy.

x/meg/sugar.go
MarcoPolo: added. Would be nice to add these to the fuzzer later

x/meg/sugar.go
MarcoPolo: Less convinced about the simpler argument, but more convinced about the idiomatic argument.

Done in 9f1651b1c2c4de91c43db6de157354afa76838a5

PR #266, opened 2025-02-12
chore: release v0.15.0

MarcoPolo:


sukunrt:
Should we have megular expressions in this?

MarcoPolo:
Hmm, that could be nice. Let me get that ready. I want to add capturing byte values first.

MarcoPolo:
On second thought, this is already going to be a tough rollout as is. Let's not make it bigger with Megular just yet. We can release Megular next when go-libp2p starts using it.

PR #269, opened 2025-02-20
feat(x/meg): Support capturing components

MarcoPolo:
Extends megular expressions to support capturing bytes as well. I have a branch that used generics, but Go is not yet good at inferring the type parameter and it looked a bit ugly. There was also not a significant performance increase in using generics for this here.

### DIFF
--- a/meg_capturers.go
+++ b/meg_capturers.go
@@ -1,0 +1,54 @@
+package multiaddr
+
+import (
+	"encoding/binary"
+	"fmt"
+	"net/netip"
+
+	"github.com/multiformats/go-multiaddr/x/meg"
+)
+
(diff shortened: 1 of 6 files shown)


sukunrt (approved):
Why not just capture the entire component?

MarcoPolo:
You're right. That would be useful for higher level capturers like "CaptureAddrPort". I'll push changes soon.

MarcoPolo:
I've exported `CaptureWithF`. That should provide what we need to build the higher level things on top of.

MarcoPolo:
I added `CaptureAddrPort` as an example capturer that composes other capturers into higher level types.

PR #271, opened 2025-02-21
refactor: keep same api as v0.14.0 for SplitFirst/SplitLast

MarcoPolo:
Perhaps not worth breaking this function, and it saves us some pain with updating downstream deps. For example, drand would not require a [backport of this change](https://github.com/drand/drand/pull/1404/files) to work with multiaddr.

The risk of then calling a method on the returned nil pointer is alleviated by changing the Component methods to be pointer receiver (as before) and checking for nil pointers explicitly.

Would also remove these breaking changes:
- https://github.com/ipfs/kubo/pull/10666/files#diff-457548d649d72a1b78a3bfea70f62723a6041821c68688be055901c4ffa141b3L810
- https://github.com/filecoin-project/lotus/pull/12912/files#diff-dd6a7924e7591cf2ed56af793a333f61f17a4a472d0ea542334ff841f5918c0dL63

### DIFF
--- a/component.go
+++ b/component.go
@@ -19,30 +19,39 @@ type Component struct {
 	valueStartIdx int // Index of the first byte of the Component's value in the bytes array
 }
 
-func (c Component) AsMultiaddr() Multiaddr {
+func (c *Component) AsMultiaddr() Multiaddr {
 	if c.Empty() {
 		return nil
 	}
-	return []Component{c}
+	return []Component{*c}
 }
 
-func (c Component) Encapsulate(o Multiaddr) Multiaddr {
+func (c *Component) Encapsulate(o Multiaddr) Multiaddr {
 	return c.AsMultiaddr().Encapsulate(o)
 }
 
-func (c Component) Decapsulate(o Multiaddr) Multiaddr {
+func (c *Component) Decapsulate(o Multiaddr) Multiaddr {
 	return c.AsMultiaddr().Decapsulate(o)
 }
 
-func (c Component) Empty() bool {
+func (c *Component) Empty() bool {
+	if c == nil {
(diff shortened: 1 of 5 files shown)


component.go
```diff
@@ -19,30 +19,39 @@ type Component struct {
 	valueStartIdx int // Index of the first byte of the Component's value in the bytes array
 }
 
-func (c Component) AsMultiaddr() Multiaddr {
+func (c *Component) AsMultiaddr() Multiaddr {
 	if c.Empty() {
 		return nil
 	}
-	return []Component{c}
+	return []Component{*c}
```
sukunrt, line 26: should make these slices with a higher capacity?

util.go
```diff
@@ -64,26 +64,26 @@ func StringCast(s string) Multiaddr {
...
-		return nil, Component{}
+		return nil, nil
 	}
 	if len(m) == 1 {
 		// We want to explicitly return a nil slice if the prefix is now empty.
-		return nil, m[0]
+		return nil, &m[0]
 	}
-	return m[:len(m)-1], m[len(m)-1]
+	return m[:len(m)-1], &m[len(m)-1]
```
sukunrt, line 86: This looks unsafe. 

last points to the last element in the slice. So appends to the first bit(the multiaddr) will modify last, right?

sukunrt: The same argument also applies to `SplitFunc`

sukunrt (approved):
Just one specific comment.

component.go
MarcoPolo: It's hard to predict how a user will use this. If they want a higher capacity they should just use the slice of Components directly

PR #275, opened 2025-03-15
feature: provide InterfaceMultiaddrsFor

tobil4sk:
net.InterfaceAddrs calls fail on android. This function allows specifying your own set of addresses, so another api can be used to avoid the issue.

See: https://github.com/multiformats/go-multiaddr/pull/255#issuecomment-2406974680

Closes #255

### DIFF
--- a/net/net.go
+++ b/net/net.go
@@ -392,7 +392,13 @@ func InterfaceMultiaddrs() ([]ma.Multiaddr, error) {
 		return nil, err
 	}
 
+	return InterfaceMultiaddrsFor(addrs)
+}
+
+// InterfaceMultiaddrsFor will return the addresses matching the given addrs
+func InterfaceMultiaddrsFor(addrs []net.Addr) ([]ma.Multiaddr, error) {
 	maddrs := make([]ma.Multiaddr, len(addrs))
+	var err error
 	for i, a := range addrs {
 		maddrs[i], err = FromNetAddr(a)
 		if err != nil {


MarcoPolo:
Is the idea that you'll override these functions with go:linkname?

tobil4sk:
The way I'm using it is to always use `InterfaceMultiaddrsFor` instead of  `InterfaceMultiaddrs`, which allows any api to be used to get the addresses.

I'm quite new to go so I'm not familiar with go:linkname. This is how I've patch a library that relies on go-multiaddr in line with this change: https://github.com/libp2p/go-libp2p/pull/3236

MarcoPolo:
Can you show me how you'll use this change? I'm having a hard time understanding how exposing `InterfaceMultiaddrsFor` helps here. Thanks

tobil4sk:
> Can you show me how you'll use this change? I'm having a hard time understanding how exposing InterfaceMultiaddrsFor helps here. Thanks

Instead of relying on `InterfaceMultiaddrs`, a slice of `net.Addr` can be passed into `InterfaceMultiaddrsFor` which allows the caller to decide how to determine the addresses. See the line:
https://github.com/libp2p/go-libp2p/pull/3236/files#diff-46c1cc328c2b33cbdce1bc123a2d6e06f5afc4758186638f9320673f8595ab0bR523

Additionally, when using `manet.ResolveUnspecifiedAddresses`, you have to avoid passing `nil` as the second argument to prevent it from calling `InterfaceMultiaddrs` internally. E.g.:
https://github.com/libp2p/go-libp2p/pull/3236/files#diff-e544d19687434c7b654de15689302596c5159444ad1db181f7096473eb42e8f3R56-R68

